PR #56, opened 2025-09-26
Fix: Convert 'assistant' role to 'model' for Google Gemini API

### LINKED ISSUE #55: [Bug]: Google Gemini uses role `model` rather than `assistant`
### Bug Description

The guide [ReqLLM guide on Context Management](https://hexdocs.pm/req_llm/data-structures.html#context-and-message-management) suggest using [`ReqLLM.Context.user/2`](https://hexdocs.pm/req_llm/ReqLLM.Context.html#user/2) and [`ReqLLM.Context.assistant/2`](https://hexdocs.pm/req_llm/ReqLLM.Context.html#assistant/2) to build the conversation history to be passed back as context to the LLM.

However Google Gemini uses `model` rather than `assistant` (passing `assistant` will return an error):
https://ai.google.dev/api#multi-turn-conversations

@mikehostetler should we add a new `ReqLLM.Context.model/2` function just for Gemini? (naming could be confusing though).

### Steps to Reproduce

Build a `ReqLLM.Context` with a message with role `:assistant`, pass it to Google Gemini and watch it crash 😅

mikehostetler:
## Summary
- Fixed issue where Google Gemini API would crash when receiving messages with role 'assistant'
- Updated `convert_messages_to_gemini` function to properly convert 'assistant' to 'model' 
- Added comprehensive tests to verify the role conversion works correctly

## Problem
The Google Gemini API expects the role `model` instead of `assistant` for AI-generated responses. When using `ReqLLM.Context.assistant/2`, the role would be encoded as `"assistant"` which causes the Gemini API to reject the request.

## Solution
Enhanced the `convert_messages_to_gemini` function in the Google provider to:
1. Handle both struct messages (with `.role`) and map messages (with `"role"` string keys)
2. Properly convert `"assistant"` and `:assistant` to `"model"`
3. Extract content from both formats correctly

## Test Plan
- [x] Added new test file `test/providers/google_role_fix_test.exs` with comprehensive tests
- [x] Verified role conversion for assistant messages
- [x] Tested system message handling
- [x] Tested different content formats (string and list)
- [x] All existing Google provider tests pass
- [x] Full test suite passes (854 tests, 0 failures)

Fixes #55

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/req_llm/providers/google.ex
+++ b/lib/req_llm/providers/google.ex
@@ -500,16 +500,37 @@ defmodule ReqLLM.Providers.Google do
   # Helper to convert OpenAI-style messages to Gemini format (non-system messages only)
   defp convert_messages_to_gemini(messages) do
     Enum.map(messages, fn message ->
+      # Extract role from either map with string keys or struct
+      raw_role =
+        case message do
+          %{role: role} -> role
+          %{"role" => role} -> role
+          _ -> "user"
+        end
+
+      # Convert role to Gemini format
       role =
-        case message.role do
+        case raw_role do
           :user -> "user"
+          "user" -> "user"
           :assistant -> "model"
-          role when is_binary(role) and role != "system" -> role
-          role when role != :system -> to_string(role)
+          "assistant" -> "model"
+          :system -> "user"  # Shouldn't happen as system messages are filtered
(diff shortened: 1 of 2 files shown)


javiercr:
Thank you!

PR #62, opened 2025-09-28
Fix Gemini decoder to keep tool calls alongside text

javiercr:
### Summary
- Gemini responses that mix free-form text with `functionCall` payloads were losing the tool invocation when decoded by our OpenAI defaults.
- Adjust the decoder to accumulate both the textual chunk and any tool call data so downstream orchestration still receives the arguments.
- Expand the Google provider contract tests with a regression case covering the mixed text + tool-call payload.

### Problem
When Gemini returns `content.parts` containing both a text entry and a `functionCall`, our default decoder hits the first clause that sees non-empty text and stops there, never emitting the tool-call chunk. As a result, `%ReqLLM.Response{}` keeps only the assistant text and `Response.tool_calls/1` comes back empty, causing orchestrators to skip the intended function execution.

### Solution
Rework the OpenAI-style message decoder to gather every chunk in the payload—text and tool calls alike—before building the response message. This ensures Gemini mixed replies preserve their tool-call metadata. Add a regression unit test in the Google provider suite to assert that text-plus-tool-call responses now return both pieces.

### Testing
```bash
mix test test/providers/google_test.exs
```

## Type of Change

- [X] Bug fix

## Testing

- [X] Tests added/updated
- [X] Manual testing performed
- [X] All tests pass

## Checklist

- [X] Code formatted and linted
- [X] Code quality checks passed (dialyzer & credo warnings/errors fixed)
- [X] Documentation updated if needed
- [X] Breaking changes documented (if any)

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -404,11 +404,7 @@ defmodule ReqLLM.Provider.Defaults do
     Enum.map(messages, &encode_openai_message/1)
   end
 
-  defp encode_openai_message(%ReqLLM.Message{
-         role: role,
-         content: content,
-         tool_calls: tool_calls
-       }) do
+  defp encode_openai_message(%ReqLLM.Message{role: role, content: content, tool_calls: tool_calls}) do
     base_message = %{
       role: to_string(role),
       content: encode_openai_content(content)
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -527,15 +527,28 @@ defmodule ReqLLM.Provider.Defaults do
 
   def default_decode_sse_event(_, _model), do: []
 
-  defp decode_openai_message(%{"content" => content}) when is_binary(content) and content != "" do
-    [ReqLLM.StreamChunk.text(content)]
+  defp decode_openai_message(%{"content" => content} = message)
+       when is_binary(content) and content != "" do
+    chunks = [ReqLLM.StreamChunk.text(content)]
+    chunks ++ decode_openai_message(Map.delete(message, "content"))
   end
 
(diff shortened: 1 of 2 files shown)


mikehostetler:
@javiercr Amazing!!! Thanks - I will look at this shortly and we can get it merged here pretty quick

mikehostetler:
@javiercr This formatting issue is weird - I hit this elsewhere and the local `mix format` doesn't fix it - I haven't had time to investigate it fully 

For now, don't worry about the pipeline and I'll make sure to test manually

javiercr:
> @javiercr This formatting issue is weird - I hit this elsewhere and the local `mix format` doesn't fix it - I haven't had time to investigate it fully
> 
> For now, don't worry about the pipeline and I'll make sure to test manually

Yeah, it's weird, formatting pass perfectly in my env.

BTW, while working on this I noticed we also need to add support for passing [functionCall](https://ai.google.dev/api/caching#FunctionCall) and [functionResponse](https://ai.google.dev/api/caching#FunctionResponse) back to the LLM as part of the context. I’m tackling that in a separate PR to enable the full flow:

image

Reference: https://ai.google.dev/gemini-api/docs/function-calling?example=meeting

One question: @mikehostetler do you know if this is already supported for other providers? I couldn’t find any relevant code for it.

Thanks!

javiercr:
> One question: @mikehostetler do you know if this is already supported for other providers? I couldn’t find any relevant code for it.

NVM, I've found:

https://github.com/agentjido/req_llm/blob/main/lib/req_llm/message/content_part.ex#L68-L74
https://github.com/agentjido/req_llm/blob/main/guides/data-structures.md#tool-result-integration

I'll figure out how to make it work for Google Gemini.

mikehostetler:
This was fixed and is now in main via #70 !

PR #72, opened 2025-10-02
Add cerebras as a direct provider and add some failsafes to defaults.ex to tackle more strict cerebras api validation

olivermt:


olivermt:
When do you re-run the models list for priv? the cerebras list is woefully outdated

mikehostetler:
I plan to run it weekly via a Github Action and auto-publish to Hex with a bump - I just need the action at this point - there's a `models_local` folder to add local patches, or you can submit a PR upstream, the same happened with the embedding models

mikehostetler:
Fixed via #78

PR #77, opened 2025-10-07
Update anthropic.json

antonioleonardo:
The context window is outdated for this particular model (there might be others with the same issue)

Are there any plans to read these values live from https://models.dev/api.json ?

Thank you!

### DIFF
--- a/priv/models_dev/anthropic.json
+++ b/priv/models_dev/anthropic.json
@@ -343,7 +343,7 @@
       "knowledge": "2025-07-31",
       "last_updated": "2025-09-29",
       "limit": {
-        "context": 1000000,
+        "context": 200000,
         "output": 64000
       },
       "modalities": {


mikehostetler:
Not planning to read them live - there's a mix task to sync it into the package so we can control the process a bit better

mikehostetler:
Did more research, the 1m token context window requires the beta header - so I'm going to leave this at 200k for now and we'll wait for models.dev to update - if we want to update this locally, it shold go into a patch file instead of the main `anthropic.json` - moving this to draft

noozo:
hey @mikehostetler, any way to pass headers to generate_text? i can see the attach_stream in anthropic provider does set some, but the prepare for text does not :(

mikehostetler:
@noozo There should be a way to pass in `req_http_options`

```
ReqLLM.generate_text(
  "anthropic:claude-3-5-sonnet",
  "Hello!",
  req_http_options: [headers: [{"x-custom-header", "value"}]]
)
```

Closing this issue for now

PR #94, opened 2025-10-11
fix(anthropic): Provider tool result encoding for multi-turn conversations

### LINKED ISSUE #93: [Bug]: Anthropic provider tool calling broken: "Unexpected role 'tool'" error
### Bug Description

The Anthropic provider in ReqLLM has two critical bugs that completely break tool/function calling functionality. When attempting to use tools in a multi-turn conversation (a non-streaming agent) with Claude, the second API request (after tool execution) fails with validation errors from Anthropic's API. This makes it impossible to use ReqLLM's tool calling feature with Anthropic models.

The root cause involves how ReqLLM's internal `Message` struct (which uses a `:tool` role for tool results) gets transformed into Anthropic's API format. ReqLLM uses a provider-specific encoder in `lib/req_llm/providers/anthropic/context.ex` to transform its generic message format into each provider's specific API format. However, this encoder has two bugs:

**Bug #1 - Invalid Role Transformation**: The `encode_message/1` function naively converts the `:tool` role atom to the string `"tool"` using `to_string(role)`. However, Anthropic's Messages API only accepts `"user"` or `"assistant"` as valid roles. According to Anthropic's documentation, tool results must be sent as `"user"` role messages containing `tool_result` content blocks, not as a separate "tool" role. When ReqLLM sends `{"role": "tool", ...}`, Anthropic's API immediately rejects the request.

**Bug #2 - Tool Result Content Not Encoded**: The `encode_content_part/1` function has a clause that matches `ContentPart` structs with `type: :tool_result`, but it simply returns `nil` instead of encoding the tool result data. This `nil` value gets filtered out by `Enum.reject(&is_nil/1)` in the `encode_content/1` function, resulting in an empty content array. Even if the role issue were fixed, Anthropic would reject messages with empty content arrays (except for the optional final assistant message).

### Steps to Reproduce

This bug manifests in any multi-turn conversation flow where Claude needs to use a tool. The first request (where Claude decides to use the tool) works fine. The bug appears on the second request when you send the tool results back to Claude for final processing.

```elixir
# In iex -S mix
alias ReqLLM.{Context, Tool}

# Step 1: Define a simple tool
tool = Tool.new!(
  name: "add",
  description: "Adds two numbers",
  parameter_schema: [
    a: [type: :integer, required: true],
    b: [type: :integer, required: true]
  ],
  callback: fn %{"a" => a, "b" => b} -> {:ok, a + b} end
)

# Step 2: Create initial conversation
messages = [
  Context.system("You are a helpful calculator assistant"),
  Context.user("What is 5 + 3?")
]

# Step 3: First API request - Claude decides to use the tool
# This works fine! Claude responds with a tool_use content block
{:ok, response1} = ReqLLM.generate_text(
  "anthropic:claude-sonnet-4-20250514",
  messages,
  tools: [tool]
)

# Step 4: Extract the tool call from Claude's response
# The response contains a ContentPart with type: :tool_call
[tool_call] = response1.message.content
  |> Enum.filter(&(&1.type == :tool_call))

tool_calls = [%{
  id: tool_call.tool_call_id,
  name: tool_call.tool_name,
  arguments: tool_call.input
}]

# Step 5: Execute the tool and append result to conversation history
# Context.execute_and_append_tools creates a Message with role: :tool
history = Context.new(messages)
history = Context.append(history, response1.message)
history = Context.execute_and_append_tools(history, tool_calls, [tool])

# At this point, history.messages contains:
# 1. %Message{role: :system, content: "You are a helpful..."}
# 2. %Message{role: :user, content: "What is 5 + 3?"}
# 3. %Message{role: :assistant, content: [%ContentPart{type: :tool_call, ...}]}
# 4. %Message{role: :tool, content: [%ContentPart{type: :tool_result, output: 8}]}

# Step 6: Second API request with tool results - THIS FAILS
{:error, error} = ReqLLM.generate_text(
  "anthropic:claude-sonnet-4-20250514",
  history.messages
)

# Error received:
# ** (RuntimeError) Anthropic API error: messages: Unexpected role "tool".
# Allowed roles are "user" or "assistant".
```

The failure occurs because the Anthropic encoder transforms message #4 into:
```json
{"role": "tool", "content": []}
```

But Anthropic expects:
```json
{"role": "user", "content": [{"type": "tool_result", "tool_use_id": "...", "content": "8"}]}
```

### Expected Behavior

Tool calling should work seamlessly end-to-end with Anthropic models, following the standard agentic flow: (1) user asks a question, (2) Claude decides to use a tool, (3) the tool executes, (4) Claude receives the result and responds to the user. This is the core value proposition of ReqLLM - providing a unified interface across different LLM providers while hiding provider-specific details.

When `Context.execute_and_append_tools/3` creates a message with `role: :tool`, the Anthropic-specific encoder should recognize this and transform it according to Anthropic's API requirements. Specifically, messages with `:tool` role containing `:tool_result` content parts should be encoded as `"user"` role messages with properly formatted `tool_result` content blocks.

The second API request should succeed and return Claude's final response.

### Actual Behavior

The tool calling flow breaks completely on the second API request. The exact error depends on which bug manifests first:

**Primary Error (Bug 1)**: When both bugs are present, the role validation fails first:
```
** (RuntimeError) Anthropic API error: messages: Unexpected role "tool".
Allowed roles are "user" or "assistant".
```

This happens because the encoder sends:
```json
{
  "role": "tool",
  "content": []
}
```

The `"tool"` role is invalid for Anthropic, and the content is empty because bug 2 caused the tool result to be encoded as `nil`.

**Secondary Error (Bug 2)**: If you hypothetically fixed just the role issue but not the content encoding, you'd get:
```
** (RuntimeError) Anthropic API error: messages.2: all messages must have
non-empty content except for the optional final assistant message
```

This happens because even with the role changed to `"user"`, the content array is still empty `[]` since the tool result wasn't properly encoded.

Both bugs must be fixed for tool calling to work. The current implementation makes it completely impossible to use ReqLLM's tool calling feature with Anthropic models - every attempt will fail on the second API request after tool execution.

### Environment

- **ReqLLM version**: 1.0.0-rc.5
- **Elixir version**: 1.18.4
- **Erlang/OTP version**: 27 [erts-15.2.7.2]
- **Provider**: Anthropic Claude (tested with claude-sonnet-4-20250514)

### Additional Context

_No response_

lostbean:
Fixes #93

## Summary

This PR resolves critical bugs preventing tool calling from working correctly
with Anthropic models in multi-turn conversations. The bugs caused tool results
to be rejected by Anthropic's API.

## The Bugs

**Bug #1: Invalid Role Transformation**

- Messages with `:tool` role were encoded as `"tool"`
- Anthropic API only accepts `"user"` or `"assistant"` roles
- Result: API error
  `"messages: Unexpected role 'tool'. Allowed roles are 'user' or 'assistant'"`

**Bug #2: Tool Result Content Not Encoded**

- `ContentPart` with `:tool_result` type returned `nil` from
  `encode_content_part/1`
- The `nil` value was filtered out by `Enum.reject(&is_nil/1)`, resulting in
  empty content
- Result: Empty tool result messages that Anthropic rejected

## Changes

**Core Fix** (`lib/req_llm/providers/anthropic/context.ex`)

- Transform `:tool` role to `:user` before encoding (line 67-77)
- Implement proper `tool_result` content block encoding (line 118-131)
- Add helper function to serialize tool outputs (strings pass through, complex
  types JSON-encoded)
- Update module documentation with correct tool result format

**Tests**

- **Provider test** (`test/providers/anthropic_test.exs`): Add multi-turn tool
  calling integration test
- **Integration test**
  (`test/req_llm/integration/tool_result_encoding_test.exs`): Add comprehensive
  provider-agnostic test suite with 7 tests covering:
  - Multi-turn conversation flow
  - Role transformation
  - Content block encoding
  - Multiple tool results
  - String vs non-string output serialization
  - Anthropic API specification compliance

## Example

```elixir
# Before: This would fail with Anthropic API errors
{:ok, response} = ReqLLM.generate_text(
  "anthropic:claude-sonnet-4-20250514",
  context_with_tool_results
)

# After: Works correctly!
{:ok, response} = ReqLLM.generate_text(
  "anthropic:claude-sonnet-4-20250514",
  context_with_tool_results
)
```

## Encoded Format (Before vs After)

**Before (Buggy):**

```elixir
%{
  role: "tool",        # ❌ Invalid role
  content: ""          # ❌ Empty content
}
```

**After (Fixed):**

```elixir
%{
  role: "user",        # ✅ Correct role
  content: [
    %{
      type: "tool_result",
      tool_use_id: "toolu_123",
      content: "8"     # ✅ Properly serialized
    }
  ]
}
```

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring
- [ ] Performance improvement
- [ ] Tests
- [ ] Build/CI

## Testing

- [x] Tests added/updated
- [ ] Manual testing performed
- [x] All tests pass

## Checklist

- [x] Code formatted and linted
- [x] Code quality checks passed (dialyzer & credo warnings/errors are
      pre-existing)
- [ ] Documentation updated if needed
- [ ] Breaking changes documented (if any)

### DIFF
--- a/lib/req_llm/providers/anthropic/context.ex
+++ b/lib/req_llm/providers/anthropic/context.ex
@@ -7,22 +7,25 @@ defmodule ReqLLM.Providers.Anthropic.Context do
   ## Key Differences from OpenAI
 
   - Uses content blocks instead of simple strings
-  - System messages are included in the messages array
-  - Tool calls are represented as content blocks with type "tool"
+  - System messages are extracted to top-level `system` parameter
+  - Tool calls are represented as content blocks with type "tool_use"
+  - Tool results must be in "user" role messages (Anthropic only accepts "user" or "assistant" roles)
   - Different parameter names (stop_sequences vs stop)
 
   ## Message Format
 
       %{
         model: "claude-3-5-sonnet-20241022",
+        system: "You are a helpful assistant",
         messages: [
-          %{role: "system", content: "You are a helpful assistant"},
-          %{role: "user", content: "Hello"},
+          %{role: "user", content: "What's the weather?"},
           %{role: "assistant", content: [
-            %{type: "text", text: "Hello!"},
-            %{type: "tool", id: "call_123", name: "weather", arguments: %{}}
+            %{type: "text", text: "I'll check that for you."},
+            %{type: "tool_use", id: "toolu_123", name: "get_weather", input: %{location: "SF"}}
           ]},
-          %{role: "tool", id: "call_123", content: "sunny"}
+          %{role: "user", content: [
+            %{type: "tool_result", tool_use_id: "toolu_123", content: "72°F and sunny"}
+          ]}
         ],
         max_tokens: 1000,
         temperature: 0.7
@@ -65,8 +68,13 @@ defmodule ReqLLM.Providers.Anthropic.Context do
   end
 
   defp encode_message(%ReqLLM.Message{role: role, content: content}) do
+    # Anthropic API only accepts "user" or "assistant" roles
+    # Tool results must be wrapped in a "user" message
+    # See: https://docs.anthropic.com/en/docs/build-with-claude/tool-use
+    normalized_role = if role == :tool, do: :user, else: role
+
     %{
-      role: to_string(role),
+      role: to_string(normalized_role),
       content: encode_content(content)
     }
   end
@@ -109,10 +117,16 @@ defmodule ReqLLM.Providers.Anthropic.Context do
 
   defp encode_content_part(%ReqLLM.Message.ContentPart{
          type: :tool_result,
-         tool_call_id: _id,
-         output: _output
+         tool_call_id: id,
+         output: output
        }) do
-    nil
+    # Anthropic tool_result format requires string content
+    # See: https://docs.anthropic.com/en/docs/build-with-claude/tool-use
+    %{
+      type: "tool_result",
+      tool_use_id: id,
+      content: serialize_tool_output(output)
+    }
   end
 
   defp encode_content_part(%ReqLLM.Message.ContentPart{
@@ -161,6 +175,10 @@ defmodule ReqLLM.Providers.Anthropic.Context do
   end
 
   defp encode_content_part(_), do: nil
+
+  # Serialize tool output to string format required by Anthropic
+  defp serialize_tool_output(output) when is_binary(output), do: output
+  defp serialize_tool_output(output), do: Jason.encode!(output)
 
   defp add_tools(request, []), do: request
 
--- a/test/providers/anthropic_test.exs
+++ b/test/providers/anthropic_test.exs
@@ -283,6 +283,79 @@ defmodule ReqLLM.Providers.AnthropicTest do
     end
   end
 
+  describe "multi-turn tool calling" do
+    test "encodes complete multi-turn conversation with tool results" do
+      # Simulate a complete tool calling flow:
+      # 1. User asks a question
+      # 2. Assistant decides to call a tool
+      # 3. Tool result is added to conversation
+      # This verifies both bugs are fixed:
+      #   - Bug #1: Tool role transformed to "user"
+      #   - Bug #2: Tool result content properly encoded
+
+      alias ReqLLM.Message.ContentPart
+
+      model = ReqLLM.Model.from!("anthropic:claude-3-5-sonnet-20241022")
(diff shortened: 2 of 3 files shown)


mikehostetler:
@lostbean Thanks!!!

PR #98, opened 2025-10-13
Refactor context and message handling for tool calls

mikehostetler:
This PR refactors the context and message handling to improve tool call support:

- Refactored Context module to support tool calls more naturally
- Added helper methods for creating assistant messages with tool calls
- Improved message handling in providers (Anthropic, Google)
- Updated examples to use the new patterns

The changes make it easier to work with tool calls in multi-turn conversations.


Valian:
@mikehostetler I really like these changes! In particular, removal of `input`, `output`,`tool_call_id` and `tool_call` from content parts. It makes everything have it's own place!

PR #99, opened 2025-10-14
chore: Update test fixtures - 2025-10-14

mikehostetler:
Updates test fixtures with latest API responses from providers.


mikehostetler:
Depends on #98

PR #111, opened 2025-10-16
Add model catalog feature for runtime discovery

mikehostetler:
## Summary
This PR adds a new model catalog feature that enables runtime discovery and querying of available AI models and providers.

## Features
- Dynamic model/provider discovery from registry files
- Filter by provider, modality (text/embedding), and capabilities
- Model detail lookup with metadata (parameters, context windows, etc.)
- Comprehensive documentation and demo examples

## Files
- `CATALOG.md` - Main documentation
- `CATALOG_DEMO.md` - Interactive demo examples

## Testing
All catalog functionality is demonstrated in the included demo file.

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -1082,27 +1082,33 @@ defmodule ReqLLM.Provider.Defaults do
     model_name = req.options[:model]
 
     # Handle case where model_name might be nil (for tests or edge cases)
-    {_provider_id, model} =
+    {_provider_id, model, model_string} =
       case model_name do
         nil ->
           # Fallback to private req_llm_model or extract from stored model
           case req.private[:req_llm_model] do
-            %ReqLLM.Model{} = stored_model -> {stored_model.provider, stored_model}
-            _ -> {:unknown, %ReqLLM.Model{provider: :unknown, model: "unknown"}}
+            %ReqLLM.Model{} = stored_model ->
+              {stored_model.provider, stored_model, stored_model.model}
+
+            _ ->
+              {:unknown, %ReqLLM.Model{provider: :unknown, model: "unknown"}, "unknown"}
           end
+
+        %ReqLLM.Model{} = model_struct ->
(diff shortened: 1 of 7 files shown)


meanderingstream:
For the catalog_demo.exs, I think we need to show more than one locally deployed model.  It will show a limitation of the existing example.  

One way of thinking about the challenge is that we are conflating the concept of a provider label with the concept of the provider behavior.  For example, if the configuration had labels, vllm and llama.cpp could map to the same provider code, lib/providers/openai.ex.  They have the same api end points for the most common situations.  If someone needs the special endpoints of vllm, then we can create a vllm.ex module.

The reason to show two models is that it demonstrates the challenge where the port for each model is different.  What if the local model was defined something like:

config = %{
      allow: %{
        vllm1: ["llama-3.2-3b-instruct"]
        vllm2: ["SmolLM2-135M-Instruct"]
      },
      overrides: [],
      custom: [
        %{
          provider: %{
            label: "vllm1",
            id: "openai"
            name: "vLLM Local 8000",
            base_url: "http://localhost:8000",
            env: ["VLLM_API_KEY"]
          },
          models: [
            %{
              id: "llama-3.2-3b-instruct",
              name: "LLaMA 3.2 3B Instruct",
              modalities: %{
                "input" => ["text"],
                "output" => ["text"]
              },
              limit: %{
                "context" => 8192
              },
              cost: %{
                "input" => 0.0,
                "output" => 0.0
              }
            }
          ],
          provider: %{
            label: "vllm2",
            id: "openai"
            name: "vLLM Local 8001",
            base_url: "http://localhost:8001",
            env: ["VLLM_API_KEY"]
          },
          models: [
            %{
              id: "SmolLM2-135M-Instruct",
              name: "SmolLM2 135M Instruct",
              modalities: %{
                "input" => ["text"],
                "output" => ["text"]
              },
              limit: %{
                "context" => 8192
              },
              cost: %{
                "input" => 0.0,
                "output" => 0.0
              }
            }
          ]
        }
      ]

PR #168, opened 2025-11-02
fix(bedrock): Exclude old Claude models from test suite

neilberkman:
Fixes test failures for old Claude models without fixtures.

Adds exclusion file to filter out Claude 3.x and legacy 4.0 models.
Only Claude 4.5 (Haiku, Sonnet) and Claude 4.1 (Opus) remain in test coverage.

Resolves issue where tests attempted to run for models lacking fixture files.


neilberkman:
Closing - Mike is refactoring to use hard-coded opt-in approach instead of exclusion lists.

PR #174, opened 2025-11-03
Add OAuth2 token caching for Google Vertex AI

neilberkman:
# Pull Request

## Description

Implements OAuth2 token caching for Google Vertex AI to eliminate 60-180ms of authentication overhead on every request. Tokens are valid for 1 hour but were being regenerated on each request, causing unnecessary file I/O, JWT signing, and HTTP round trips.

The cache uses a GenServer + ETS pattern with a 55-minute TTL (5-minute safety margin before expiry). The GenServer serializes concurrent refresh requests to prevent duplicate token fetches.

## Type of Contribution

- [x] **Core Library** - Changes to core modules or data structures
- [ ] **New Provider** - Adding a new LLM provider
- [x] **Provider Feature** - Adding capabilities to existing provider
- [ ] **Bug Fix** - Fixing existing functionality
- [ ] **Documentation** - Docs/guides only

## Checklist

- [x] Tests pass (`mix test`)
- [x] Documentation updated

### If Provider Changes
- [ ] Fixtures generated (`mix mc "provider:*" --record`)
- [ ] Model compatibility passes (`mix mc "provider:*"`)

**Model Compatibility Output:**
```
# N/A - No API changes to provider

```

## Related Issues

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Support for both Google AI Studio and Vertex AI (requires Gemini models)
   - `cached_content` provider option to reference existing caches
   - Minimum token requirements: 1,024 (Flash) / 4,096 (Pro)
+- **OAuth2 token caching for Google Vertex AI**
+  - Eliminates 60-180ms auth overhead on every request
+  - Tokens cached for 55 minutes (5 minute safety margin before 1 hour expiry)
+  - GenServer serializes concurrent refresh requests to prevent duplicate fetches
+  - Per-node cache (no distributed coordination needed)
+  - 99.9% reduction in auth overhead for typical workloads
 - **Real-time stream processing** with `ReqLLM.StreamResponse.process_stream/2`
   - Process streams incrementally with real-time callbacks
   - `on_result` callback for content chunks (fires immediately as text arrives)
--- a/lib/req_llm/application.ex
+++ b/lib/req_llm/application.ex
@@ -20,7 +20,8 @@ defmodule ReqLLM.Application do
     children =
       [
         {Finch, finch_config},
-        {Task.Supervisor, name: ReqLLM.TaskSupervisor}
+        {Task.Supervisor, name: ReqLLM.TaskSupervisor},
+        ReqLLM.Providers.GoogleVertex.TokenCache
       ] ++ dev_children()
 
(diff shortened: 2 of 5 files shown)


neilberkman:
## Security Analysis

Wanted to document security considerations for caching OAuth2 tokens in memory:

### What's Being Cached

This caches OAuth2 **access tokens** (not API keys or service account credentials). Important distinctions:

- **Access tokens**: Short-lived bearer tokens (1 hour validity, cached 55 min)
- **API keys**: Long-lived permanent credentials (different security profile)
- **Service account JSON**: Permanent credentials (never cached here, only file path stored)

### Memory Storage Risks

**Potential exposure vectors:**
1. **Memory dumps** - If an attacker gains access to dump process memory (via debugging tools, core dumps, etc.), cached tokens could be extracted
2. **Process inspection** - Erlang observer/tracing tools could potentially view ETS contents

**Mitigations in this implementation:**
- Table is `:private` (not `:public`) - only the owning GenServer process can access it
- Tokens expire after 55 minutes, limiting exposure window
- GenServer supervision ensures table cleanup on crash
- No permanent credentials stored (only file paths)

### Comparison: Public vs Private ETS

The codebase has two ETS patterns:

**Schema cache (`:public :named_table`)** - Different security profile:
```elixir
:ets.new(:req_llm_schema_cache, [:public, :named_table, ...])
:ets.lookup(:req_llm_schema_cache, key)  # Any process can access
```

**Token cache (`:private`)** - This PR:
```elixir
table = :ets.new(@table_name, [:private, ...])  # Only GenServer can access
:ets.lookup(state.table, key)
```

The `:private` pattern provides better isolation since other processes cannot directly read the table.

### Security Tradeoff

**Risk**: Tokens in memory for up to 55 minutes vs storing in ephemeral memory for milliseconds per request

**Benefit**: Eliminates 60-180ms overhead per request (JWT signing, HTTP round trip to Google)

**Alternative approaches if this is deemed too risky:**
1. No caching (current main branch) - fetch token every request
2. Filesystem caching with restricted permissions - moves exposure from memory to disk
3. External cache service (Redis, etc.) - adds deployment complexity

### Recommendation

For library code running in trusted environments (typical Elixir deployments), the risk seems acceptable given:
- Short token lifetime (55 min)
- Private table isolation
- Significant performance benefit (99.9% reduction in auth overhead)

However, if users deploy in environments with heightened security requirements (multi-tenant, untrusted code, etc.), they may prefer the no-cache approach despite the performance cost.

Worth documenting this tradeoff in module docs?

PR #241, opened 2025-11-27
Fixes an issue where finish_reason wasn't captured when streaming

zblanco:
Handles scenarios where finish_reason is bundled instead of exclusive and includes reasoning fields in extra option keys.

### Changes
1. Updated ReqLLM.StreamServer (lib/req_llm/stream_server.ex):

* Modified enqueue_chunks/2 to merge ALL metadata from stream chunks into the server state, not just usage. This ensures that finish_reason (and potentially other provider-specific metadata) is captured and available in the final response.
* Updated ReqLLM.Provider.Defaults (lib/req_llm/provider/defaults.ex):

2. Rewrote default_decode_stream_event/2 to correctly handle SSE events that contain BOTH content/choices AND usage metadata in the same chunk.
* Previously, the pattern matching was exclusive, prioritizing chunks with usage and ignoring finish_reason inside choices within the same chunk.
* The new implementation iterates over choices to extract content and finish_reason, AND separately extracts usage, combining them into a list of chunks.

## Type of Contribution

- [x] **Bug Fix** - Fixing existing functionality

## Checklist

- [x] Tests pass (`mix test`)

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -390,7 +390,9 @@ defmodule ReqLLM.Provider.Defaults do
         :presence_penalty,
         :seed,
         :stop,
-        :user
+        :user,
+        :reasoning_effort,
+        :reasoning_token_budget
       ] ++ provider_mod.supported_provider_options()
 
     {api_key, extra_option_keys}


zblanco:
Most of the local failures running `mix mc "*:*"` are for new models without generated fixtures. The reasoning fields fix surfaced after changes to merging options so tests pass and previously failing compat tests pass but new models fail - you might want to run them on your end in case I missed something.

PR #245, opened 2025-11-29
Add Azure OpenAI provider (#127)

### LINKED ISSUE #127: [Feature]: Azure configuration
### Feature Type

Other

### Provider (if applicable)

azure

### Description

When working with our models hosted on azure, the easiest path has been to use openai provider, but override certain parameters. The `req_http_options` have been great for it, as it lets us add custom headers for their non-standard `api-key` header auth on older versions, and override the end url. 

On the latest `main` branch, the ability to override the resulting URL was removed as it always uses `base_url <> path()`. 

There are two problems to solve when connecting to models hosted on azure specifically when dealing with the responses type api:
- Azure requires us to specify an `api-version` url parameter. The current workaround is to set the base_url to a path that ends in `#` so that the rest of the url is ignored, and thus we can manually specify the api-version parameter by overriding the `base_url`.
- Azure requires the `deployment_name` instead of the actual model name in the request body. We go around this by editing the model field in model struct.

Here's some sample code:

```elixir
  def stream_text(text) do
    {:ok, model} = ReqLLM.model("openai:gpt-5-mini")

    model
    |> Map.put(:model, "gpt-5-mini-2025-08-07")
    |> ReqLLM.stream_text(text,
      base_url:
        "https://#{endpoint_url}/openai/responses?api-version=2025-04-01-preview#"
    )
  end
```

Starting a discussion on how to best support these extra fields that azure needs (`endpoint_url`, `deployment_name`, `api_version`). I see a couple ways to implement:

- Pass these fields as provider_opts, implement an azure provider which extends openai provider and uses the fields into the url. The con here is that one would need to pass these options to every call. 
- Let users build their own AzureProvider and add these fields to the DSL. This would be ergonomic to use, but I had a lot of trouble implementing a custom provider when using req_llm as a library. 
- add them as allowed options when creating a model with the `azure` provider, use the fields if present to build the url at the provider level. This is a good tradeoff, could be the way forward.

Please let me know your preferred option so we can start working on a PR to add it. 

### Example Usage

_No response_

### Can you contribute this?

Yes, I can submit a PR

shelvick:
## Summary

- Add Azure AI provider supporting OpenAI and Anthropic models hosted on Azure
- Fix Responses API tool format for strict mode (flat format + additionalProperties)
- Extract shared adapter helpers for OpenAI and Anthropic providers
- Improve streaming response handling and usage parsing

### Supported Models

**OpenAI family:** GPT-4o, GPT-4.1, GPT-5, GPT-5.1, o1/o3/o4 reasoning, codex

**Anthropic family:** Claude 4.5 (Opus, Sonnet, Haiku), Claude Opus 4.1

### Features

- Deployment-based routing with custom Azure endpoints
- API key authentication via `api-key` header
- Family-specific environment variable resolution
- Full streaming support with usage tracking
- Tool calling and structured output
- Extended thinking for Claude and reasoning models

Closes #127

## Test plan

- [x] All existing tests pass (`mix test`)
- [x] Azure provider tests pass (`mix test test/provider/azure/`)
- [x] Azure coverage tests pass with fixtures (`mix test test/coverage/azure/`)


shelvick:
Some fixtures are missing because I couldn't deploy those models. All of those are either too old, or Microsoft just won't give me access because I'm a pleb, lol.

mikehostetler (commented):
## Review Summary

Great PR adding Azure AI provider support! I reviewed the code and ran tests - everything works well. I found one issue that needs addressing:

### Issue: Dialyzer Warning in Azure.OpenAI.parse_response/3

The current implementation creates temporary `Req.Request` structs to delegate to `default_decode_response`, which triggers a dialyzer type mismatch warning. The fix is to use `Defaults.decode_response_body_openai_format/2` directly instead.

### Suggested Fix

Replace the `parse_response/3` function in `lib/req_llm/providers/azure/openai.ex` with:

```elixir
@doc """
Parses an Azure OpenAI response into ReqLLM format.

Uses the centralized OpenAI response decoding from Provider.Defaults.
"""
def parse_response(body, model, opts) do
  context = opts[:context] || %ReqLLM.Context{messages: []}
  operation = opts[:operation]

  {:ok, response} = Defaults.decode_response_body_openai_format(body, model)

  merged_response = ReqLLM.Context.merge_response(context, response)

  final_response =
    if operation == :object do
      extract_and_set_object(merged_response)
    else
      merged_response
    end

  {:ok, final_response}
end

defp extract_and_set_object(response) do
  extracted_object =
    response
    |> ReqLLM.Response.tool_calls()
    |> ReqLLM.ToolCall.find_args("structured_output")

  %{response | object: extracted_object}
end
```

### Also: Formatting Fix

`lib/req_llm/stream_response.ex` needs a minor formatting fix at line 784-785 (run `mix format`).

### Tests

All 1640 tests pass and `mix quality` passes after these fixes.

mikehostetler:
I've pushed fixes to address the CI failures and merge conflicts:

1. **Resolved merge conflicts** with main branch (PR #247 tool encoding fix)
2. **Fixed dialyzer warning** in `Azure.OpenAI.parse_response/3` - now uses `Defaults.decode_response_body_openai_format/2` directly instead of creating temporary Req structs
3. **Applied formatting fixes** to `stream_response.ex`

All 1640 tests pass and `mix quality` passes locally. The PR should be ready to merge once CI completes.

PR #35, opened 2025-09-21
Fix #31: Add cost calculation to Response.usage()

mikehostetler:
## Summary

This PR fixes issue #31 by implementing cost calculation in the ReqLLM.Response.usage/1 function and adds comprehensive support for OpenAI's cached tokens pricing.

### Core Implementation
- **Updated Response struct**: Changed usage field type from %{optional(atom()) => integer()} to map() to support cost fields
- **Enhanced Step.Usage**: Extended to handle %ReqLLM.Response{} bodies and inject cost breakdown into Response.usage
- **New cost breakdown logic**: Replaced compute_cost/2 with compute_cost_breakdown/2 that calculates:
  - input_cost: Cost for input tokens (with cached/uncached split for OpenAI)
  - output_cost: Cost for output tokens  
  - total_cost: Sum of input and output costs

### OpenAI Cached Tokens Support 🆕
- **Extended Model cost type**: Now supports optional cached_input rate for 50% discount pricing
- **Cached token extraction**: Extracts cached_tokens from prompt_tokens_details in OpenAI responses
- **Smart cost calculation**: Splits input tokens into cached vs uncached and applies different rates
- **Response integration**: Exposes cached_tokens in Response.usage() when present
- **Edge case handling**: Proper clamping, type conversion, and validation

### Key Features
- **Backward compatibility**: Existing token fields (input_tokens, output_tokens, total_tokens) preserved
- **Cost fields added**: When model has pricing data, usage map includes input_cost, output_cost, total_cost
- **Cached token support**: When OpenAI provides cached tokens, proper discount pricing applied
- **Graceful degradation**: Cost fields only added when pricing is available and tokens are valid numbers
- **Response preservation**: All other Response fields maintained when adding cost breakdown

### Testing
- Added comprehensive test coverage for Response struct cost calculation
- Added extensive cached token tests covering OpenAI scenarios
- Tests handle edge cases: malformed tokens, missing cost data, graceful failures
- All existing tests continue to pass (backward compatibility verified)

## Before/After

**Before:**
```elixir
{:ok, response} = ReqLLM.generate_text("openai:gpt-4", "Hello")
ReqLLM.Response.usage(response)
# => %{input_tokens: 8, output_tokens: 9, total_tokens: 17}
```

**After (with cost calculation):**
```elixir
{:ok, response} = ReqLLM.generate_text("openai:gpt-4", "Hello")  
ReqLLM.Response.usage(response)
[...]

PR #253, opened 2025-12-01
feat: Add Azure OpenAI embedding model support

shelvick:
## Summary

- Enable embedding generation for Azure OpenAI models (text-embedding-3-small, text-embedding-3-large, text-embedding-ada-002)
- Fix `decode_response` to return raw body for embedding operations, matching the OpenAI provider's behavior
- Add test fixtures for all three embedding models

## Dependencies

This PR depends on [agentjido/llm_db#34](https://github.com/agentjido/llm_db/pull/34) which adds the Azure embedding model definitions to LLMDB.

## Test plan

- [x] Recorded fixtures for all three Azure embedding models
- [x] Verified embedding tests pass with `mix mc "azure:text-embedding-*" --type embedding`

PR #254, opened 2025-12-02
feat: Add Azure DeepSeek model support

shelvick:
## Summary

Add support for 5 Azure DeepSeek models:
- `azure:deepseek-v3.1` - DeepSeek V3.1 with tool calling
- `azure:deepseek-v3-0324` - DeepSeek V3 March 2024 with tool calling
- `azure:deepseek-r1` - DeepSeek R1 reasoning model
- `azure:deepseek-r1-0528` - DeepSeek R1 May 2028 reasoning model
- `azure:mai-ds-r1` - Microsoft AI's DeepSeek R1 (uncensored variant)

### Key Changes

- **Model family routing**: Added "deepseek" and "mai-ds" prefixes to Azure provider, routing to OpenAI-compatible formatter
- **Environment variables**: Family-specific env vars (`AZURE_DEEPSEEK_*`, `AZURE_MAI_*`) with fallback to `AZURE_*`
- **Reasoning token inference**: Fixed usage parsing to detect `reasoning_content` field (Azure DeepSeek R1 returns reasoning here but doesn't populate `completion_tokens_details.reasoning_tokens`)
- **Forced tool choice**: Azure DeepSeek doesn't support object-format tool_choice (`{type: "function", function: {name: "..."}}`), only string values ("auto", "required", "none"). Tests now check `tools.forced_choice` capability.

### Dependencies

- **Requires [llm_db PR #35](https://github.com/agentjido/llm_db/pull/35)** which adds the `tools.forced_choice` capability field

## Test plan

- [x] All 1695 tests pass
- [x] `mix quality` passes
- [x] Tested against live Azure DeepSeek endpoints with fixture recording
- [x] Verified reasoning token inference works for R1 models
- [x] Verified tool calling works for V3 models with `tool_choice: "required"`


test/support/provider_test/comprehensive.ex
```diff
@@ -488,7 +503,7 @@ defmodule ReqLLM.ProviderTest.Comprehensive do
               {:ok, response} =
                 ReqLLM.generate_object(
                   @model_spec,
-                  "Generate a software engineer profile",
+                  "Generate a fictional software engineer profile with a made-up name and age",
```
mikehostetler: @shelvick this change is a deal breaker, as it will force redoing all of our fixture tests

mikehostetler:
I'm closing #253 as this PR has superseded it - I need the changes to `comprehensive.ex` backed out before I can merge this though - which will unfortunately break all of your fixtures - sorry - then happy to get it landed

shelvick:
> I'm closing #253 as this PR has superseded it - I need the changes to `comprehensive.ex` backed out before I can merge this though - which will unfortunately break all of your fixtures - sorry - then happy to get it landed

Yeah sorry, I think I accidentally based these changes on top of the embedding ones when they could/should have been separate.

Did the forced_choice stuff break other tests? Is that why you need it reversed?

mikehostetler:
I will do some testing on the `forced_choice` - I am not as concerned about that. The problem is changing the default prompts - because that breaks fixtures for the 125+ working models

shelvick:
@mikehostetler It's reverted.

FYI, I did change those fixtures along with the prompt, so the tests were still passing. I made that change because a couple of the models were giving me lip about the prompt being too vague when I was trying to record new fixtures for them. Though I understand if you just didn't want to make that invasive of a change -- it's a minor thing, not super likely to happen again.

mikehostetler:
Thanks! I'll take a look!

PR #281, opened 2025-12-12
feat(anthropic): Add message caching support for conversation prefixes

shelvick:
## Summary

- Adds `anthropic_cache_messages` option to enable caching of conversation history
- When enabled alongside `anthropic_prompt_cache: true`, adds a cache breakpoint at the last message
- Subsequent requests with identical messages up to the breakpoint are served from cache, reducing costs and latency

## Use Case

Multi-turn conversations where the same context is sent repeatedly. Cache hits can reduce costs by up to 90%.

```elixir
ReqLLM.chat(model, context,
  provider_options: [
    anthropic_prompt_cache: true,
    anthropic_cache_messages: true,
    anthropic_prompt_cache_ttl: "1h"  # optional
  ]
)
```

## Implementation

- Added `anthropic_cache_messages` boolean to `@provider_schema`
- Added `maybe_cache_last_message/3` to the existing caching pipeline (after tools and system)
- Adds `cache_control` to last content block of last message
- Preserves existing `cache_control` on content blocks
- Uses same TTL as system/tools via `anthropic_prompt_cache_ttl`

## Test plan

- [x] Adds cache_control to last message when enabled
- [x] Adds cache_control to last content block of multi-part messages
- [x] Applies TTL correctly
- [x] Does nothing when `anthropic_cache_messages: false`
- [x] Does nothing when `anthropic_prompt_cache: false`
- [x] Handles empty messages gracefully
- [x] Works alongside system and tool caching
- [x] Preserves existing cache_control on content blocks


neilberkman:
I appreciate the fixes!

PR #303, opened 2025-12-29
feat: Add thinking parameter support for Z.ai providers

georgeguimaraes:
## Summary

- Add support for the `thinking` parameter in both `zai` and `zai_coder` providers
- Users can control the thinking/reasoning mode via `provider_options`
- Setting `type: "disabled"` provides faster responses by skipping the reasoning phase
- Setting `type: "enabled"` enables extended reasoning for more complex tasks

## Usage

```elixir
ReqLLM.generate_text("zai:glm-4.5-flash", context,
  provider_options: [thinking: %{type: "disabled"}]
)
```

## Changes

- Updated `zai.ex` to pass the `thinking` parameter from `provider_options` to the API request body
- Updated `zai_coder.ex` with the same change
- Added documentation for the new provider option in both module docs

## References

- [Z.ai API Documentation on thinking parameter](https://bigmodel.cn/dev/api/normal-model/glm-4)

### DIFF
--- a/lib/req_llm/providers/zai.ex
+++ b/lib/req_llm/providers/zai.ex
@@ -23,6 +23,20 @@ defmodule ReqLLM.Providers.Zai do
 
       # Add to .env file (automatically loaded)
       ZAI_API_KEY=your-api-key
+
+  ## Provider Options
+
+  The following options can be passed via `provider_options`:
+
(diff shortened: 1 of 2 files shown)


mikehostetler:
Thanks @georgeguimaraes !

PR #360, opened 2026-01-26
feat(req): Allows setting up Req Plugins on High-level api

### LINKED ISSUE #359: [Feature]: Allow setting up Req Attachments without the need of using low level API
### Feature Type

Core library enhancement

### Provider (if applicable)

_No response_

### Description

Of course, we can always use the low-level API, but this would need to reimplement the whole high-level API just to add REQ plugins, so lets just allow REQ Plugins to be set!

### Example Usage

```elixir
{:ok, req} = ReqLLM.generate_text("yada yada", req_plugins: [&MyPlugin.attach/1])
```

### Can you contribute this?

Yes, I can submit a PR

ycastorium:
## Description

Allows setting up Req Plugins on High-level api

## Type of Change

- [x] Documentation update

## Testing

- [x] Tests pass (`mix test`)

## Checklist

## Related Issues

Closes #359

### DIFF
--- a/lib/req_llm/providers/azure.ex
+++ b/lib/req_llm/providers/azure.ex
@@ -525,7 +525,7 @@ defmodule ReqLLM.Providers.Azure do
       end)
     end)
     |> Req.Request.register_options(extra_option_keys ++ [:deployment])
-    |> Req.Request.merge_options(user_opts)
+    |> Req.Request.merge_options(ReqLLM.Provider.Defaults.filter_req_opts(user_opts))
     |> ReqLLM.Step.Retry.attach()
     |> ReqLLM.Step.Error.attach()
     |> Req.Request.append_response_steps(llm_decode_response: &decode_response/1)


mikehostetler:
Thanks for the contribution! However, I'm going to hold off on merging this for now:

1. **Finch streaming not supported** - The `:req_plugins` option only works for non-streaming operations. Streaming uses Finch directly via `attach_stream/4`, so plugins wouldn't apply there. This creates an inconsistent API surface.

2. **Moving away from Req** - We're planning to move away from Req in the future, which would make this feature short-lived and potentially a breaking change down the road.

I appreciate the effort here, but given these concerns I don't think this is the right addition at this time.

ycastorium:
> Thanks for the contribution! However, I'm going to hold off on merging this for now:
> 
>     1. **Finch streaming not supported** - The `:req_plugins` option only works for non-streaming operations. Streaming uses Finch directly via `attach_stream/4`, so plugins wouldn't apply there. This creates an inconsistent API surface.
> 
>     2. **Moving away from Req** - We're planning to move away from Req in the future, which would make this feature short-lived and potentially a breaking change down the road.
> 
> 
> I appreciate the effort here, but given these concerns I don't think this is the right addition at this time.

Aha! Good to know, that you are planning that! Thanks for the input.

mikehostetler:
Thanks again for the original contribution. I'm adding `needs_work` to reflect current blocker status before this can land.

This PR is still draft and the current direction concerns from maintainers remain: streaming parity and long-term API direction around Req plugin exposure in high-level APIs.

If you want to continue this, please update scope/proposal to address those points and we can re-evaluate.

mikehostetler:
Thanks again for the contribution here. We're going to close this for the current version rather than keep it open in `needs_work`.

The main blockers are still the same ones we called out earlier:
- `:req_plugins` only applies to the non-streaming Req path today, while streaming goes through Finch, so the API surface would stay inconsistent.
- We're also moving away from exposing Req-specific behavior at the high-level API, so this would likely be short-lived and hard to support cleanly.

The implementation work itself is appreciated, but this isn't a direction we want to take in this release. If the long-term transport/plugin story changes later, we can revisit the idea from a fresh proposal.

PR #412, opened 2026-02-11
feat: enable reasoning support for DeepSeek models on Azure

shelvick:
## Summary

- Add `deepseek_model?/1` helper to `AdapterHelpers` for detecting DeepSeek model IDs
- Stop stripping `thinking` config from `additional_model_request_fields` for DeepSeek models in `pre_validate_options` (it was being removed with an "Anthropic-specific" warning, but DeepSeek supports it natively)
- Default to `thinking: %{type: "enabled"}` in `format_request` for DeepSeek models, enabling reasoning by default
- Users can override via `additional_model_request_fields` (e.g., custom `budget_tokens`) or suppress entirely by setting `thinking: nil`
- Fix model ID extraction in `pre_validate_options` to use `provider_model_id` when available, matching `effective_model_id/1` used elsewhere in the Azure module

## Notes

- DeepSeek R1 models always reason regardless of the `thinking` parameter — confirmed via live test that sending it is harmless (ignored, no error)
- `mai-ds-r1` (Microsoft's DeepSeek R1 derivative) uses the `"mai-ds"` model family prefix and is correctly not matched by `deepseek_model?/1`

## Test plan

- 10 new tests covering `deepseek_model?/1`, `pre_validate_options` thinking preservation, `format_request` default injection, user override, opt-out via nil, and regression checks for non-DeepSeek models
- Full suite: 2196 tests, 0 failures

### DIFF
--- a/lib/req_llm/providers/azure/openai.ex
+++ b/lib/req_llm/providers/azure/openai.ex
@@ -94,7 +94,6 @@ defmodule ReqLLM.Providers.Azure.OpenAI do
   end
 
   defp warn_and_remove_anthropic_thinking_config(opts, model_id) do
-    # DeepSeek models support thinking config natively; don't strip it
     if AdapterHelpers.deepseek_model?(model_id) do
       opts
     else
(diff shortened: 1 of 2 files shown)


shelvick:
@mikehostetler This was fixed too, I just didn't leave a comment. (Is the comment needed? Sorry, I'm bad at GitHub.)

mikehostetler:
@shelvick You're all good - I sw th e commits - I'll work these through shortly!

PR #484, opened 2026-03-09
fix: handle list-format content in convert_loose_map for thinking models

mykewould:
Thinking models (e.g. Qwen 3.5 via Ollama) return assistant messages with list-format content containing thinking and text parts. The existing convert_loose_map only handled binary content, causing {:error, :invalid_loose_map} when conversation history included assistant responses from thinking models.

Add a new convert_loose_map clause that matches %{role: atom, content: list} and extracts text parts while stripping thinking parts.

I'll add more tests, docs, etc if there's no major issues with the overall concept.

### DIFF
--- a/lib/req_llm/context.ex
+++ b/lib/req_llm/context.ex
@@ -714,6 +714,10 @@ defmodule ReqLLM.Context do
     end
   end
 
+  defp to_parts(list) when is_list(list) do
+    Enum.flat_map(list, &to_part_list/1)
+  end
+
   defp normalize_tool_calls(nil), do: nil
   defp normalize_tool_calls([]), do: nil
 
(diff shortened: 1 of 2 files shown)


mikehostetler:
Thanks for the fix here. I re-ran the branch locally and there are still two blocking issues before this is safe to merge:

1. `extract_text_from_content_parts/1` does not actually filter real `ContentPart.thinking/1` structs. `ReqLLM.Message.ContentPart.thinking/1` stores the payload in `text`, not `thinking`, so the current match misses it and reasoning text gets replayed into the next assistant message. Repro on this branch:

```elixir
ReqLLM.Context.normalize(%{
  role: :assistant,
  content: [
    ReqLLM.Message.ContentPart.thinking("internal"),
    ReqLLM.Message.ContentPart.text("hello")
  ]
})
#=> {:ok, #Context<1 msgs: assistant:"internalhello">}
```

2. The new loose-map path still does not handle JSON-decoded list-format content. It only matches atom-key maps, so OpenAI-style histories like this still fail:

```elixir
ReqLLM.Context.normalize(%{
  "role" => "assistant",
  "content" => [%{"type" => "text", "text" => "answer"}]
})
#=> {:error, :invalid_loose_map}
```

I also ran `mix test test/providers/google_test.exs test/req_llm/context_test.exs` on the branch and it passed, so this looks like a coverage gap rather than a broader breakage. Once those two cases are handled, this should be in much better shape.

mikehostetler:
Thanks for the fix here. I'm going to close this one out as superseded rather than leave it in `needs_work`.

I re-checked the branch against current `main`, and the functional change from this PR is already present upstream now. The `Context` normalization logic and tests from this branch match what has already landed on `main`, so there isn't a remaining diff here to merge.

Appreciate the contribution and the follow-up on this issue.

Issue #259, opened 2025-12-03
[Feature]: Support video_url for multimodal messages


genblazejohn:
### Feature Type

Core library enhancement

### Description

The Problem

Right now, req_llm lets us send text and images (using image_url), but there is no built-in way to send a video URL.

Since models can  understand video, we need a standard way to pass a video link in our messages list without writing custom upload code every time.

Proposed Solution:

I propose adding a video_url type to the message list, similar to how image_url works.

If the provider supports it (e.g., Gemini): The library should just pass the URL through to the provider's native parameter (like fileData).

If the provider doesn't support it: The library should just raise an error saying "Video URLs are not supported for this provider." It does not need to handle downloading or processing the video file itself.

Why we need this

This keeps the API consistent. Users can send text, images, and videos using the same structure.

### Example Usage

```elixir
messages = [
  %{
    role: "user",
    content: [
      %{"type" => "text", "text" => "What is happening in this video?"},
      %{"type" => "video_url", "video_url" => %{"url" => "[https://example.com/video.mp4](https://example.com/video.mp4)"}}
    ]
  }
]

ReqLLM.chat(messages, model: "gemini-1.5-pro")
```

### Can you contribute this?

Yes, with guidance

PR #556, opened 2026-03-24
feat: add video_url multimodal support

mikehostetler:
## Summary
- add a first-class `ContentPart.video_url/1,2` API and normalize raw `video_url` message maps
- support `video_url` URLs on Google Gemini by translating them to Gemini `fileData` inputs
- raise a clear unsupported-provider error on the default OpenAI-style encoding path and document the new content part

## Testing
- mix test test/req_llm/message/content_part_test.exs test/req_llm/context_test.exs test/req_llm/provider/defaults_test.exs test/providers/google_test.exs
- mix compile --warnings-as-errors
- mix quality

Refs #259

### DIFF
--- a/guides/data-structures.md
+++ b/guides/data-structures.md
@@ -118,6 +118,8 @@ Typed content elements that compose a `Message`. Common variants:
 - `text/2`: `ContentPart.text("...", metadata)` with metadata map
 - `image_url/1`: `ContentPart.image_url("https://...")`
 - `image_url/2`: `ContentPart.image_url("https://...", metadata)` with metadata
+- `video_url/1`: `ContentPart.video_url("https://...")`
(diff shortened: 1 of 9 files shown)


mikehostetler:
Thanks for the follow-up here. I'm marking this one `needs_work` because it's currently `DIRTY` against `main`, and the overlapping `video_url` implementation in #549 is the cleaner review target right now. I didn't find a material behavior advantage here that offsets the merge conflict risk, so my recommendation is to keep #549 as the survivor and only bring over any extra doc polish from this branch if we still want it.

mikehostetler:
Closing in favor of #549, which is the cleaner surviving implementation for issue #259. I compared the two directly: #556 adds a small docs touch and some refactoring, but it does not provide a materially better implementation than #549, and it is still conflict-dirty against main.

PR #558, opened 2026-03-24
feat: add litellm provider

mikehostetler:
## Summary
- add a LiteLLM provider built on the existing OpenAI-compatible defaults
- support `litellm:model` string specs via inline fallback when the model is not in the catalog
- document the local proxy default, auth expectations, and model-spec workflow

## Testing
- mix test test/providers/litellm_test.exs test/req_llm_test.exs test/req_llm/providers_test.exs test/req_llm/availability_test.exs
- mix docs
- mix quality


mikehostetler:
Closing this per current repo direction. We are not taking the LiteLLM provider path forward right now.

PR #561, opened 2026-03-24
feat: allow unknown models with warnings instead of errors

### LINKED ISSUE #358: [Feature]: Relax strict model verification to support unknown/new models and proxies
### Feature Type

Core library enhancement

### Provider (if applicable)

_No response_

### Description

current implementation of ReqLLM.model/1 imposes strict validation against the local LLMDB catalog. While this ensures type safety and capability checking for known models, it creates significant friction in several common real-world scenarios.

The Problem If a user attempts to use a valid model spec string (e.g., provider:new-model) that is not yet present in the local llm_db snapshot, the library raises a validation error and blocks the request. This effectively prevents users from using any model that hasn’t been explicitly indexed by the library maintainers.

Key Use Cases Affected

New Model Releases: LLM providers release new models frequently. There is often a lag between a model’s release and the corresponding update to req_llm/llm_db. Users cannot use the new model immediately without waiting for a library update or manually monkey-patching. 

API Aggregators: Services like OpenRouter, Zenmux, or others often support a vast array of models, including those from smaller providers or fine-tunes that may never be in the core catalog. 

Reverse Proxies / Custom APIs: Users running self-hosted proxies or standardizing via OpenAI-compatible endpoints often map models to custom names (e.g., my-internal-mix) which will fail current validation.

### Example Usage

_No response_

### Can you contribute this?

Yes, I can submit a PR

dl-alexandre:
Closes #358

## Problem

The current implementation of ReqLLM.model/1 imposes strict validation against the local LLMDB catalog. When a user attempts to use a model not in the catalog (new releases, API aggregators like OpenRouter, custom proxies), the library raises an error and blocks the request.

## Solution

Modified the model fallback logic in resolve_provider_model_fallback/3 to:

1. Check if the provider is valid/available (exists and is configured)
2. If valid, create an inline model and emit a helpful warning about unverified models
3. If invalid, raise a clear error about the unknown provider

## Changes

- Extended resolve_provider_model_fallback/3 to handle all providers, not just :openai_codex
- Added descriptive warning when using unverified models (includes provider:model ID and guidance)
- Maintains backward compatibility - catalog models work exactly as before

## Benefits

- Users can immediately use newly released models (no waiting for LLMDB updates)
- API aggregators like OpenRouter work seamlessly with their vast model catalogs
- Custom/proxy endpoints with non-standard model names are supported
- Helpful warnings inform users about potential limitations (pricing, token counting, capabilities)
- Users can suppress warnings by using explicit inline model specs

### DIFF
--- a/.agents/skills/pr-harden/SKILL.md
+++ b/.agents/skills/pr-harden/SKILL.md
@@ -1,0 +1,79 @@
+---
+name: pr-harden
+description: Take a pull request labeled needs_work, address blocking review findings, strengthen test coverage, rerun validation, and flip it to ready_to_merge when it is clean.
+---
+
+# PR Harden
+
+Use this skill when a PR is labeled `needs_work` and the goal is to make the PR merge-ready.
+
+Review target: <url>
+
+## Goal
+
+Turn a `needs_work` PR into a `ready_to_merge` PR by fixing the blocking issues directly on the PR branch.
+
+## Workflow
+
+1. Start from a clean, current `main`
+   - `git status --short --branch`
+   - `git fetch --prune origin`
+   - `git checkout main`
+   - `git pull --ff-only origin main`
+
+2. Inspect the PR before changing code
+   - Read the PR description, files, current checks, merge state, comments, and reviews
+   - Identify the blocking issues first: correctness, regressions, missing tests, red CI, merge conflicts
+   - Use the existing review labels as truth: this skill is for PRs labeled `needs_work`
+
+3. Check out the PR branch for editing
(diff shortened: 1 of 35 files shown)


mikehostetler:
Review triage marked this PR as `needs_work`.

Blocking issue:
- The PR currently carries an unrelated streaming retry change where the 429 retry path in `lib/req_llm/streaming/retry.ex` passes `state.headers` into `maybe_retry/11` instead of the full state map. That breaks rate-limit retry classification and drops `Retry-After` handling for streaming 429s.

I would rebase or split this PR down to the model-resolution change, then add focused regression coverage for the new unknown-model fallback behavior before relabeling it ready to merge.

PR #578, opened 2026-04-06
Fix: preserve inference profile prefix in Bedrock API URLs

### LINKED ISSUE #579: [Bug]: Bedrock inference profile prefix stripped from API URLs
**Provider:** `amazon_bedrock`

**Model:** `global.anthropic.claude-opus-4-6-v1`

## Bug Description

When using Bedrock models with inference profile prefixes (e.g., `global.`, `us.`, `eu.`), the prefix is stripped during LLMDB catalog lookup but never restored for API URL construction. The Bedrock API URL uses the stripped model ID (`/model/anthropic.claude-opus-4-6-v1/invoke`) instead of the prefixed one (`/model/global.anthropic.claude-opus-4-6-v1/invoke`).

### History and context

The Bedrock provider and inference profile support were built up in a series of PRs:

- **#66** (Oct 2025, @neilberkman) — Original Bedrock provider with streaming support
- **#104** (Oct 2025, @neilberkman) — Added `normalize_model_id` callback for Bedrock inference profiles. At this point, llm_db preserved the full prefixed ID (e.g., `global.anthropic.claude-opus-4-6-v1`) in `model.id`, so the prefix was naturally available for URL construction.
- **#152** (Nov 2025, @neilberkman) — Added `preserve_inference_profile?` callback on the Anthropic formatter, designed as forward-looking infrastructure for the case where the prefix might be stripped during lookup. Since llm_db 2026.2.x preserved the prefix, the callback wasn't needed in the request path yet.
- **#449** (Feb 2026, @stevehodgkiss) — Expanded the region prefix list to include `au`, `apac`, `jp`, etc. for Australian inference profiles.

Starting with **llm_db 2026.3.x**, `normalize_model_id` began stripping inference profile prefixes for catalog matching. This is correct for metadata lookup, but the stripped ID is now the only one available downstream — `model.id` loses the prefix and `model.provider_model_id` is nil. The `preserve_inference_profile?` callback was designed for exactly this scenario but was never wired into the request path because it was never needed while llm_db preserved the prefix.

The `attach` function at `amazon_bedrock.ex:358` uses `model.provider_model_id || model.id` for the API URL — both are stripped, so the prefix is lost.

Models that have on-demand access (e.g., Sonnet 4.5) happen to work without the prefix because Bedrock falls back to on-demand in the configured region. Models that only support cross-region inference profiles (e.g., Opus 4.6) fail.

## Reproduction Code

```elixir
# With llm_db >= 2026.3.x
{:ok, model} = ReqLLM.model("amazon_bedrock:global.anthropic.claude-opus-4-6-v1")
model.id
#=> "anthropic.claude-opus-4-6-v1"  (global. prefix stripped)
model.provider_model_id
#=> nil

# API URL is built as: /model/anthropic.claude-opus-4-6-v1/invoke
# Should be: /model/global.anthropic.claude-opus-4-6-v1/invoke

{:ok, response} = ReqLLM.generate_text(
  "amazon_bedrock:global.anthropic.claude-opus-4-6-v1",
  "Hello",
  api_key: "...",
  max_tokens: 100
)
```

## Expected Behavior

The `global.` prefix should be preserved in the Bedrock API URL. Models accessed via cross-region inference profiles should work regardless of on-demand availability.

## Actual Behavior

```
** (ReqLLM.Error.API.Request) Provider response error (400): Bedrock API error:
Invocation of model ID anthropic.claude-opus-4-6-v1 with on-demand throughput
isn't supported. Retry your request with the ID or ARN of an inference profile
that contains this model.
```

## Environment

- ReqLLM: 1.9.0
- llm_db: 2026.3.3 / 2026.4.0 (both affected)
- Elixir: 1.19.5
- OTP: 28

Fix submitted in #578.

neilberkman:
## Description

When using Bedrock models with inference profile prefixes (e.g., `global.`, `us.`, `eu.`), the prefix is stripped during LLMDB catalog lookup but never restored for API URL construction. Models like `global.anthropic.claude-opus-4-6-v1` that only support cross-region inference profiles (not on-demand access) fail with: "Invocation of model ID anthropic.claude-opus-4-6-v1 with on-demand throughput isn't supported."

This wasn't an issue with earlier llm_db versions (e.g., 2026.2.8) which preserved the prefix in the model struct. Starting with llm_db 2026.3.x, `normalize_model_id` strips inference profile prefixes for catalog matching but doesn't preserve the original prefixed ID anywhere on the returned model struct. See #579 for full history.

This fix adds `maybe_restore_inference_prefix/2` in `ReqLLM.model/1` which detects when a prefix was stripped during catalog lookup and restores it as `provider_model_id`. Since all Bedrock URL construction already uses `model.provider_model_id || model.id`, this fixes all code paths (non-streaming, streaming, embedding) without changing the provider code.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None. Models without inference profile prefixes continue to work identically.

## Testing

- [x] Tests pass (`mix test`)

## Checklist

## Related Issues

Closes #579

### DIFF
--- a/lib/req_llm.ex
+++ b/lib/req_llm.ex
@@ -330,7 +330,9 @@ defmodule ReqLLM do
   def model(spec) when is_binary(spec) do
     case LLMDB.model(spec) do
       {:ok, %LLMDB.Model{} = model} ->
-        normalize_model_result({:ok, model})
+        {:ok, model}
+        |> maybe_restore_inference_prefix(spec)
+        |> normalize_model_result()
 
       {:error, _reason} = error ->
         spec
@@ -391,6 +393,49 @@ defmodule ReqLLM do
     do: {:ok, normalize_model_metadata(model)}
 
   defp normalize_model_result(other), do: other
+
+  # Bedrock cross-region inference profile prefixes.
+  # These are stripped by LLMDB during catalog lookup but must be preserved
+  # in API URLs for models that only support inference profiles (not on-demand).
+  @bedrock_inference_prefixes ~w(us eu ap apac ca au jp us-gov global)
+
+  # When a Bedrock model is resolved from a spec like "amazon_bedrock:global.anthropic.claude-opus-4-6-v1",
+  # the LLMDB catalog lookup strips the inference profile prefix for matching.
+  # This function detects that specific case and restores the original prefixed
+  # model ID as provider_model_id so it's used in API URL construction.
+  #
+  # Scoped to Bedrock inference profile prefixes only — does NOT affect alias
+  # resolution for other providers (e.g., anthropic:claude-3-haiku resolving
+  # to claude-3-haiku-20240307 should NOT set provider_model_id).
+  defp maybe_restore_inference_prefix(
+         {:ok, %LLMDB.Model{provider: :amazon_bedrock} = model},
+         spec
+       )
+       when is_binary(spec) do
+    original_model_id = extract_model_id_from_spec(spec)
+
+    if original_model_id && has_inference_prefix?(original_model_id) do
+      {:ok, %{model | provider_model_id: original_model_id}}
+    else
+      {:ok, model}
+    end
+  end
+
+  defp maybe_restore_inference_prefix({:ok, %LLMDB.Model{}} = result, _spec), do: result
+
+  defp has_inference_prefix?(model_id) do
+    case String.split(model_id, ".", parts: 2) do
+      [prefix, _rest] -> prefix in @bedrock_inference_prefixes
+      _ -> false
+    end
+  end
+
+  defp extract_model_id_from_spec(spec) do
+    case String.split(spec, ":", parts: 2) do
+      [_provider, model_id] -> model_id
+      _ -> nil
+    end
+  end
 
   defp normalize_inline_model_result({:ok, %LLMDB.Model{} = model}, _attrs) do
     {:ok, normalize_model_metadata(model)}
--- a/lib/req_llm/providers/amazon_bedrock.ex
+++ b/lib/req_llm/providers/amazon_bedrock.ex
@@ -354,7 +354,9 @@ defmodule ReqLLM.Providers.AmazonBedrock do
     base_url = "https://bedrock-runtime.#{region}.amazonaws.com"
 
     # Use provider_model_id if set (for models requiring specific API format like inference profiles),
-    # otherwise fall back to canonical model ID
+    # otherwise fall back to canonical model ID.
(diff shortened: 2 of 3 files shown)


mikehostetler (dismissed):
The Bedrock bug is real, but the new  logic in  is too broad. It restores the original spec into  for any provider whenever the requested model ID differs from the catalog ID. That means normal alias resolution like  would now produce  even though the resolved catalog ID is , and providers prefer uid=501(mhostetler) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),79(_appserverusr),80(admin),81(_appserveradm),701(com.apple.sharepoint.group.1),33(_appstore),98(_lpadmin),100(_lpoperator),204(_developer),250(_analyticsusers),395(com.apple.access_ftp),398(com.apple.access_screensharing),399(com.apple.access_ssh),400(com.apple.access_remote_ae) when building API requests. So this PR would turn non-Bedrock aliases into outbound API model IDs too. I’d scope the restoration to Bedrock’s stripped inference-profile prefixes specifically, or otherwise only preserve the known Bedrock prefix case rather than any string mismatch.

mikehostetler (changes requested):
The Bedrock bug is real, but the new maybe_restore_inference_prefix/2 logic in ReqLLM.model/1 is too broad. It restores the original spec into provider_model_id for any provider whenever the requested model ID differs from the catalog ID. That means normal alias resolution like anthropic:claude-3-haiku would now produce provider_model_id "claude-3-haiku" even though the resolved catalog ID is claude-3-haiku-20240307, and providers prefer provider_model_id || id when building API requests. So this PR would turn non-Bedrock aliases into outbound API model IDs too. I’d scope the restoration to Bedrock’s stripped inference-profile prefixes specifically, or otherwise only preserve the known Bedrock prefix case rather than any string mismatch.

neilberkman:
Good catch, thanks. Updated — `maybe_restore_inference_prefix/2` now pattern-matches on `provider: :amazon_bedrock` and checks that the model ID starts with a known inference profile prefix (`global.`, `us.`, `eu.`, `ap.`, `ca.`, `au.`, `jp.`, `us-gov.`). Non-Bedrock alias resolution (e.g., `anthropic:claude-3-haiku` → `claude-3-haiku-20240307`) is not affected. Added a test verifying that case.

mikehostetler (approved):
The Bedrock prefix restoration is now scoped correctly to the inference-profile case, the regression concern from the earlier review is addressed, and CI is green. This looks merge-ready.

PR #586, opened 2026-04-10
fix(anthropic): handle malformed tool arguments gracefully

BlueHotDog:
## Problem

`decode_tool_arguments/1` in the Anthropic context encoder uses `Jason.decode!/1`, which raises when the LLM sends an empty string `""` as tool arguments. This happens with parameter-free tools (e.g. `take_screenshot()`) where some models send `""` instead of `"{}"` or `null`.

## Fix

Replace `Jason.decode!/1` with `ReqLLM.JSON.decode/1` (already used elsewhere in the codebase, e.g. `ToolCall.parse_arguments/2`) so malformed or empty arguments fall back to `%{}` instead of crashing.

```elixir
# before
defp decode_tool_arguments(args) when is_binary(args), do: Jason.decode!(args)

# after
defp decode_tool_arguments(args) when is_binary(args) do
  case ReqLLM.JSON.decode(args) do
    {:ok, map} -> map
    {:error, _} -> %{}
  end
end
```

## Test

Added a test case that encodes a conversation history containing a `ToolCall` with `arguments: ""` and asserts the encoded `input` is `%{}`.

### DIFF
--- a/lib/req_llm/providers/anthropic/context.ex
+++ b/lib/req_llm/providers/anthropic/context.ex
@@ -267,7 +267,13 @@ defmodule ReqLLM.Providers.Anthropic.Context do
     %{type: "tool_use", id: id, name: name, input: decode_tool_arguments(args)}
   end
 
-  defp decode_tool_arguments(args) when is_binary(args), do: Jason.decode!(args)
+  defp decode_tool_arguments(args) when is_binary(args) do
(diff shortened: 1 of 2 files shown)


mikehostetler (approved):
Reviewed the current head against main, checked focused local validation, and the current CI is green. This looks merge-ready.

PR #594, opened 2026-04-10
feat(openai): add explicit Ollama auth marker

### LINKED ISSUE #583: [Feature]: Make unauthenticated Ollama support explicit for OpenAI-compatible models
### Feature Type

Provider feature extension

### Provider (if applicable)

Ollama

## Problem

`ReqLLM` allows users to route OpenAI-compatible backends through the OpenAI provider by passing inline `%LLMDB.Model{}` specifications with a custom base_url.

That works well for wire compatibility, but it creates an authentication ambiguity.

Today, an inline model such as:

```elixir
ReqLLM.model!(%{
  provider: :openai,
  id: "llama3",
  base_url: "http://localhost:11434/v1"
})
```

is still just an `%LLMDB.Model{provider: :openai}` at runtime. From the provider’s perspective, this does not distinguish:

- OpenAI itself
- Ollama
- local OpenAI-compatible proxies
- remote OpenAI-compatible services

As a consequence, the OpenAI provider cannot reliably decide whether API-key authentication is required without relying on `base_url` heuristics.

## Why This Matters

Ollama’s default local endpoint does not require an API key, while the OpenAI provider does by default.

A URL-based exception is possible, but it is not a particularly good boundary:

- it encodes product semantics through transport details
- it is difficult to make obviously correct
- it does not scale well to other OpenAI-compatible backends
- it risks silently relaxing authentication for endpoints that should remain authenticated

## Proposal

Make this distinction explicit in model or provider metadata instead of inferring it from the URL.

One possible direction would be to support inline model metadata such as:

```elixir
ReqLLM.model!(%{
  provider: :openai,
  id: "llama3",
  base_url: "http://localhost:11434/v1",
  extra: %{
    openai_compatible_backend: :ollama
  }
})
```

or an equivalent provider option.

Then the OpenAI provider can apply a narrow rule:

- if the backend is explicitly :ollama, missing API-key auth may be allowed
- otherwise, current OpenAI authentication behavior remains unchanged

This keeps the behavior explicit, local to the OpenAI provider, and independent from URL guessing.

## Alternative

A stronger but larger alternative would be to introduce a first-class :ollama provider that reuses the OpenAI wire format internally.

That would give the cleanest semantic boundary, but it also increases maintenance surface and may be unnecessary if the only difference is authentication and small compatibility handling.

## Open Questions

- Should the discriminator live in model metadata, provider options, or both?
- Should this be strictly Ollama-specific, or should we design a more general OpenAI-compatible backend marker from the start?
- If we introduce an explicit discriminator, do we want to preserve any temporary URL-based fallback for backwards compatibility?
- Would a dedicated `:ollama` provider be preferable if more Ollama-specific behavior is expected over time?

## Acceptance Criteria

- Explicitly marked Ollama models can be used without `OPENAI_API_KEY`
- Unmarked OpenAI-compatible endpoints continue to require authentication by default
- OAuth validation continues to behave as it does today
- Tests cover both the Ollama case and the guard cases for non-Ollama endpoints

## Notes

The core issue is not authentication itself, but the lack of a reliable structural signal in `%LLMDB.Model{provider: :openai}` to express
which OpenAI-compatible backend is being targeted. Any solution that makes that distinction explicit should improve both correctness and maintainability.

### Example Usage

_No response_

### Can you contribute this?

Yes, I can submit a PR

mikehostetler:
## Summary
- add an explicit `openai_compatible_backend: :ollama` marker for OpenAI-compatible backends via `provider_options` or `model.extra`
- omit bearer auth only for explicitly marked Ollama requests when no OpenAI API key is configured, while preserving the default auth requirement and existing OAuth behavior
- cover the new behavior across text generation, streaming request building, and transcription, and update the inline localhost OpenAI example to show the explicit marker

## Validation
- `mix test test/req_llm/generation_test.exs test/req_llm/transcription_test.exs test/req_llm/base_url_streaming_test.exs test/req_llm/openai_websocket_test.exs test/req_llm/openai_compatible_backend_test.exs test/provider/openai`
- `mix quality`

Fixes #583

### DIFF
--- a/lib/req_llm.ex
+++ b/lib/req_llm.ex
@@ -39,7 +39,8 @@ defmodule ReqLLM do
         ReqLLM.model!(%{
           provider: :openai,
           id: "gpt-6-mini",
-          base_url: "http://localhost:8000/v1"
+          base_url: "http://localhost:8000/v1",
+          extra: %{openai_compatible_backend: :ollama}
         })
 
       ReqLLM.generate_text(model, messages)
--- a/lib/req_llm/providers/openai.ex
+++ b/lib/req_llm/providers/openai.ex
@@ -192,6 +192,11 @@ defmodule ReqLLM.Providers.OpenAI do
       doc:
         "Streaming transport for Responses models. Use :websocket for OpenAI WebSocket mode; SSE remains the default."
     ],
+    openai_compatible_backend: [
+      type: {:or, [{:in, [:ollama]}, {:in, ["ollama"]}]},
+      doc:
+        "Explicit OpenAI-compatible backend marker. Set to :ollama to allow missing API keys for unauthenticated local deployments."
+    ],
     tool_outputs: [
       type: {:list, :any},
       doc:
@@ -439,7 +444,7 @@ defmodule ReqLLM.Providers.OpenAI do
         |> maybe_add_transcription_part(:language, language)
         |> maybe_add_transcription_provider_parts(provider_options)
 
-      api_key = ReqLLM.Keys.get!(model, opts)
+      credential = resolve_request_credential!(model, opts)
 
       request =
         Req.new(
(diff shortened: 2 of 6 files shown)


ccarvalho-eng:
Thanks for closing this issue!

PR #601, opened 2026-04-13
fix: preserve capabilities when model/1 receives a populated LLMDB.Model struct

mhsdef:
## Description

When `ReqLLM.model/1` receives a fully-populated `%LLMDB.Model{}` struct (e.g. constructed via `LLMDB.Model.new!/1` with explicit capabilities), it was tearing the struct apart, stripping nil fields, running it through `LLMDB.Enrich.enrich_model/1`, and rebuilding via `LLMDB.Model.new()`. This caused caller-provided fields like `capabilities` to be overwritten by Zoi schema defaults (e.g. `json.schema: false`, `tools.enabled: false`).

This matters because the recommended advanced usage path — constructing an explicit `LLMDB.Model` struct for models not yet in the catalog or with incorrect catalog data — was silently losing the capabilities the caller set. This caused `determine_output_mode/2` to fall back to `:tool_strict` instead of using native `:json_schema` structured output, even when the caller explicitly declared `json: %{schema: true}`.

**Root cause:** `LLMDB.Enrich.enrich_model/1` only derives `family` and `provider_model_id` — it doesn't do catalog lookups. After enrichment, the original non-nil values were being discarded. The fix merges enriched fields (family, provider_model_id) back *under* the caller's original values, so explicit capabilities are never overwritten.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None. The existing behavior for bare structs (e.g. `LLMDB.Model.new!(%{id: "foo", provider: :openai})` with no capabilities) is unchanged — enrichment still fills in `family` and `provider_model_id`.

## Testing

- [x] Tests pass (`mix test`) — 2814 tests, 0 failures
- [x] New test: verifies capabilities survive the `model/1` round-trip for a fully-populated struct
- [x] Existing enrichment test still passes (bare struct gets family/provider_model_id derived)

## Checklist

## Related Issues

Discovered while investigating why `determine_output_mode/2` was returning `:tool_strict` for models with `json.schema: true` capabilities — the capabilities were being silently dropped during `model/1` normalization.

### DIFF
--- a/lib/req_llm.ex
+++ b/lib/req_llm.ex
@@ -288,11 +288,15 @@ defmodule ReqLLM do
   """
   @spec model(model_input()) :: {:ok, LLMDB.Model.t()} | {:error, term()}
   def model(%LLMDB.Model{} = model) do
-    model
-    |> Map.from_struct()
-    |> Enum.reject(fn {_key, value} -> is_nil(value) end)
-    |> Map.new()
-    |> LLMDB.Enrich.enrich_model()
+    non_nil =
+      model
+      |> Map.from_struct()
+      |> Enum.reject(fn {_key, value} -> is_nil(value) end)
+      |> Map.new()
+
+    enriched = LLMDB.Enrich.enrich_model(non_nil)
+
+    Map.merge(enriched, non_nil)
     |> LLMDB.Model.new()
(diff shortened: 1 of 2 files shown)


mhsdef:
Nope.

PR #618, opened 2026-04-17
fix(openai): dedupe Responses API commentary and final answer

zblanco:
We found a really small edge case (that I'm frankly surprised we noticed at all) where GPT-5.4 sometimes produces a commentary and final_answer that is the same so when concatenated it looks like a duplicate output.

## Description

Fixes duplicate text in OpenAI Responses API output when phased `message` segments include identical `commentary` and `final_answer` text. Adds targeted unit tests covering both the duplicate phased-output case and the case where commentary differs from the final answer, which should remain preserved.

## Type of Change

## Breaking Changes

None.

## Testing

- [x] Tests pass (`mix test`)

Ran focused validation: `mix test test/provider/openai/responses_api_unit_test.exs`

## Checklist

## Related Issues

N/A

### DIFF
--- a/lib/req_llm/providers/openai/responses_api.ex
+++ b/lib/req_llm/providers/openai/responses_api.ex
@@ -1341,6 +1341,7 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
       |> Enum.filter(&(&1["type"] in ["output_text", "text"]))
       |> Enum.map(&extract_text_field/1)
     end)
+    |> Enum.uniq()
     |> Enum.join("")
     |> case do
       "" -> nil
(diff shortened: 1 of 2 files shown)


mikehostetler (changes requested):
Thanks for chasing this down. There is still a blocking regression here: applying Enum.uniq/1 after flattening every message text fragment removes legitimate repeated output, not just duplicated commentary/final-answer pairs. I reproduced it locally with message fragments ["ha", "ha", "!"], which now decode to "ha!" instead of "haha!". The dedupe needs to happen at the duplicated message level or be gated on phase/content equality so repeated assistant text is preserved.

PR #623, opened 2026-04-19
feat(azure): support Azure OpenAI v1 GA API (no api-version)

neilberkman:
## Description

Adds support for Azure OpenAI's [v1 GA API](https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-lifecycle) in the existing `:azure` provider. The v1 GA API (launched August 2025) drops the dated `api-version` query parameter and uses `/openai/v1/...` URL paths with the model/deployment name in the request body — matching standard OpenAI shape.

Detection is automatic from the base URL: when the path contains `/openai/v1`, chat/responses/embedding requests use the GA paths and skip the `api-version` query string. No new options or breaking changes — existing dated/Foundry users are unaffected.

| Endpoint   | Dated                                                                         | v1 GA                |
|------------|-------------------------------------------------------------------------------|----------------------|
| chat       | `/deployments/{deployment}/chat/completions?api-version=...`                  | `/chat/completions`  |
| responses  | `/responses?api-version=...`                                                  | `/responses`         |
| embeddings | `/deployments/{deployment}/embeddings?api-version=...`                        | `/embeddings`        |

For v1 GA, the deployment name is added to the request body's `model` field (same mechanism as the existing Foundry path).

## Type of Change

## Breaking Changes

None. Existing `:azure` users on dated `api-version` URLs and Foundry URLs see no change in behavior.

## Testing

- [x] Tests pass (`mix test`) — 101 Azure tests, 0 failures
- [x] Quality checks pass (`mix quality`) — format, compile, dialyzer (0 errors), credo --strict
- [x] Live-tested end-to-end against a real Azure resource (`gpt-5.3-chat` deployment, `https://*.cognitiveservices.azure.com/openai/v1` base URL): request hits `/chat/completions` with no `api-version`, model name in body, returns 200 with usage tokens.

New unit tests cover:
- `uses_v1_ga_format?/1` detection (positive, negative, edge cases)
- v1 GA chat URL path + no `api-version`
- v1 GA embedding URL path + no `api-version`
- v1 GA streaming URL path + no `api-version`
- Model name added to request body for v1 GA

## Checklist

- [x] I have updated the documentation accordingly (moduledoc usage example for v1 GA)
- [x] I have **NOT** edited \`CHANGELOG.md\` (it is auto-generated by git_ops)

## Related Issues

N/A — opened directly as a feature contribution. Happy to file a tracking issue if preferred.

### DIFF
--- a/lib/req_llm/providers/azure.ex
+++ b/lib/req_llm/providers/azure.ex
@@ -88,6 +88,17 @@ defmodule ReqLLM.Providers.Azure do
         deployment: "deepseek-v3"
       )
 
+      # Azure OpenAI v1 GA API (auto-detected from /openai/v1 path).
+      # No api-version parameter; deployment name is sent in the request body.
+      # See https://learn.microsoft.com/azure/ai-services/openai/api-version-lifecycle
+      ReqLLM.generate_text(
+        "azure:gpt-5.3-chat",
+        "Hello!",
+        api_key: "your-api-key",
+        base_url: "https://my-resource.openai.azure.com/openai/v1",
+        deployment: "gpt-5.3-chat"
+      )
+
   ## Examples
 
       # Basic usage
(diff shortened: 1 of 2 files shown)


mikehostetler:
One small edge case I hit while reviewing: the new v1 detection explicitly accepts a trailing-slash base URL like "https://.../openai/v1/", but attach_stream/4 still builds the streaming URL with raw string concatenation. Reproducing locally with that documented base URL shape produced "/openai/v1//responses", so streaming requests end up with a double-slash path in that case.

Everything else I checked in the touched paths looked good, and the focused tests passed on my side. I think this just needs base URL normalization (or a URI-aware join) plus a regression test for the trailing-slash form before it is merge-clean.

neilberkman:
Good catch, thanks for reviewing.

Fixed in 671bc86:

- Added a small `join_url/2` helper that strips trailing slash from base URL before appending the path
- `attach_stream/4` now uses it instead of raw string interpolation
- New regression test asserts the streaming path doesn't contain `//` when the caller passes the trailing-slash form documented in the moduledoc (`https://...openai/v1/`)

102 tests pass, `mix quality` clean.

PR #624, opened 2026-04-20
fix(openai): encode multimodal tool outputs in Responses API

BlueHotDog:
## Summary
- route inline `:tool` messages in Responses API request building through a shared encoder
- emit array `function_call_output.output` when tool result content contains image parts (`input_text` + `input_image` items)
- preserve existing string output behavior for non-multimodal tool results
- add unit coverage for a text+image tool result to ensure array output is produced

## Testing
- `mix test test/provider/openai/responses_api_unit_test.exs`

### DIFF
--- a/lib/req_llm/providers/openai/responses_api.ex
+++ b/lib/req_llm/providers/openai/responses_api.ex
@@ -519,24 +519,7 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
             # most recent round's outputs were appended, which caused
             # "No tool output found for function call" errors on multi-turn
             # tool calling with the Responses API.
-            output =
-              case ReqLLM.ToolResult.output_from_message(msg) do
-                nil -> extract_tool_output_text(msg.content)
-                value -> value
-              end
-
-            output_string =
-              cond do
-                is_binary(output) -> output
-                is_map(output) or is_list(output) -> Jason.encode!(output)
-                true -> to_string(output)
-              end
-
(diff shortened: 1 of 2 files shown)


mikehostetler (changes requested):
Thanks for the fix here. The image path looks right, but this still leaves file-bearing tool results broken in the Responses API. `function_call_output.output` can be an array of text, image, or file items, yet `encode_tool_message_inline/1` only switches to array output when `has_image_content?(msg.content)` is true. A tool result like `[text("Attached report"), file(...)]` still falls back to `extract_tool_output_text/1` and is serialized as just `"Attached report"`, dropping the file content entirely. I reproduced that locally on this PR head. This needs to promote file content to array output as well and add a regression test for that case.

BlueHotDog:
Addressed in fe36b51.

- `encode_tool_message_inline/1` now treats `ContentPart` `:file` as multimodal, so file-bearing tool results are emitted as array `function_call_output.output` instead of collapsing to text.
- Added regression test `encodes file-bearing tool results as array function_call_output` in `test/provider/openai/responses_api_unit_test.exs`.

Validation: `mix test test/provider/openai/responses_api_unit_test.exs` (91 tests, 0 failures).

mikehostetler (approved):
Thanks for the follow-up here. I reran the focused unit suite and rechecked the text+file tool-result path locally on this head; the file part is now preserved in the `function_call_output.output` array, so the blocking issue from the previous review is resolved.

PR #636, opened 2026-04-21
fix(openai): preserve responses assistant phases

### LINKED ISSUE #634: Responses API: round-trip assistant phase for GPT-5 phase-aware replay
## Summary
`ReqLLM.Providers.OpenAI.ResponsesAPI` currently **does not round-trip assistant `phase`** (`commentary` / `final_answer`) through context replay.

This makes phase-aware replay impossible in manual replay flows (without `previous_response_id`) and can degrade long-running GPT-5/Codex tool trajectories.

## Why This Matters
OpenAI’s GPT-5 guidance recommends preserving assistant `phase` in Responses API replay for long-running/tool-heavy flows.

- Docs: https://developers.openai.com/api/docs/guides/latest-model ("New phase parameter")
- Cookbook: https://cookbook.openai.com/examples/gpt-5/codex_prompting_guide

When `phase` is dropped, prior commentary can be treated as final output, increasing early-stop / wrong-turn behavior.

### Decode path
In `lib/req_llm/providers/openai/responses_api.ex`, decode logic handles phased output only for text aggregation/dedup, then drops phase metadata:

- `dedupe_phased_messages/1` exists and is used during text aggregation.
- `decode_responses_success/1` builds assistant message metadata with only `response_id`:
  - `metadata: %{response_id: body["id"]}`

So phase info is not preserved in `ReqLLM.Message` / `Context`.

### Encode path
In the same module, assistant messages are encoded into `input` items without `phase`:

- assistant branch emits `%{"role" => "assistant", "content" => ...}`
- no support for `metadata[:phase]` or multi-phase assistant replay items.

## Repro (conceptual)
```elixir
alias ReqLLM.{Context}
alias ReqLLM.Providers.OpenAI.ResponsesAPI

ctx = Context.new([
  Context.assistant("I’ll inspect logs", metadata: %{phase: "commentary", response_id: "resp_1"}),
  Context.assistant("Root cause found", metadata: %{phase: "final_answer", response_id: "resp_1"}),
  Context.user("continue")
])

body = ResponsesAPI.build_request_body(ctx, "gpt-5.4", %{stream: false, provider_options: []}, nil)

# Expected: assistant input items include "phase"
# Actual: phase is absent
```

## Impact
- Lower continuation quality in manual replay flows for GPT-5 Responses workloads.
- More retry churn / tool-loop instability on long trajectories.
- Lost observability fidelity for intermediate vs final assistant output.

This is especially relevant in integrations that intentionally avoid `previous_response_id` / `store: true` and rely on full manual replay.

### 1) Preserve phase on decode
In `decode_responses_success/1`, extract phase info from output message segments and preserve it in assistant metadata.

Potential metadata shape:
- `metadata[:phase]` for single-phase message
- `metadata[:phase_items]` (ordered) for multi-segment phase output

### 2) Emit phase on encode
In `build_request_body/4`, assistant encoding should:
- include `"phase"` when metadata has `phase`
- if `phase_items` exists, emit ordered assistant input items with corresponding phases

### 3) Keep compatibility
- Ignore unknown/invalid phase values.
- Do not emit phase for user/system/tool roles.
- Preserve current behavior when no phase metadata is present.

## Suggested Tests
In `test/provider/openai/responses_api_unit_test.exs`:
1. decode preserves phase metadata from output message segments.
2. encode includes `phase` when assistant metadata has phase.
3. round-trip test: phased output -> decoded context -> re-encoded input retains phase.
4. behavior remains unchanged for non-phased responses.

## Notes
There are already tests for deduping commentary/final text, but no tests that verify phase preservation/replay in request input.

mikehostetler:
## Summary
- preserve assistant `phase` metadata when decoding Responses API output and streamed completion metadata
- replay ordered phased assistant items during manual history encoding, while still carrying tool calls forward
- add regression coverage for decode, stream metadata propagation, replay encoding, and round-trip behavior

## Validation
- mix test test/provider/openai/responses_api_unit_test.exs
- mix test

Closes #634

### DIFF
--- a/lib/req_llm/provider/defaults/response_builder.ex
+++ b/lib/req_llm/provider/defaults/response_builder.ex
@@ -338,7 +338,18 @@ defmodule ReqLLM.Provider.Defaults.ResponseBuilder do
         base
       end
 
-    base
+    base =
+      if metadata[:phase] do
+        Map.put(base, :phase, metadata[:phase])
+      else
+        base
+      end
+
+    if metadata[:phase_items] do
+      Map.put(base, :phase_items, metadata[:phase_items])
+    else
+      base
+    end
   end
 
   defp generate_response_id do
--- a/lib/req_llm/providers/openai/responses_api.ex
+++ b/lib/req_llm/providers/openai/responses_api.ex
@@ -74,6 +74,7 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
     "mcp_call" => :mcp_call,
     "x_search_call" => :x_search_call
   }
+  @assistant_phases ["commentary", "final_answer"]
 
   @impl true
   def path, do: "/responses"
@@ -207,6 +208,7 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
   defp capture_completion_metadata(data, meta) do
     usage_data = get_in(data, ["response", "usage"])
     response_id = get_in(data, ["response", "id"])
+    response_output = get_in(data, ["response", "output"]) || []
 
     meta =
       if response_id do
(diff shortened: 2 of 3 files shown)


BlueHotDog:
ship it :)

PR #646, opened 2026-04-25
fix: encode thinking content parts as reasoning_content for OpenAI-compatible providers

spfoos:
## Problem
Moonshot, DeepSeek, and other OpenAI-compatible providers that support thinking/reasoning return reasoning_content as a top-level message field.

ReqLLM correctly decodes this into ContentPart{type: :thinking} during response parsing, but when encoding messages back to the API request, the thinking parts were dropped (encode_openai_content_part returned nil).

This caused multi-turn conversations with reasoning models to fail with:
> thinking is enabled but reasoning_content is missing in assistant tool call message

## Fix
Extract thinking content parts during message encoding, join their text into a reasoning_content string, and place it at the top level of the message object while removing thinking parts from the content array.

## Backwards Compatibility
Non-reasoning messages are unaffected (maybe_add_field skips nil values). Providers that do not use reasoning_content simply ignore the extra field.

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -836,9 +836,21 @@ defmodule ReqLLM.Provider.Defaults do
          reasoning_details: rd,
          metadata: metadata
        }) do
+    {thinking_texts, content_without_thinking} =
+      Enum.split_with(c, fn
+        %ReqLLM.Message.ContentPart{type: :thinking} -> true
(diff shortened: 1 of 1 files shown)


spfoos:
Closed — the bug was actually in ash_ai, not req_llm. req_llm correctly surfaces and serializes thinking content. See https://github.com/ash-project/ash_ai/pull/194 for the correct fix.

PR #647, opened 2026-04-25
fix: encode thinking content parts as reasoning_content for OpenAI-compatible providers

spfoos:
When assistant messages contain ContentPart.thinking(...) parts (e.g. from Moonshot reasoning models), the OpenAI encoder was stripping them via encode_openai_content_part returning nil for :thinking parts.

This caused multi-turn conversations with Moonshot to fail with:
> thinking is enabled but reasoning_content is missing in assistant tool call message

Changes:
- Add extract_reasoning_content/1 to pull thinking parts out of message content
- In encode_openai_message, extract reasoning_content before encoding content, then add it as a top-level reasoning_content field on the encoded message
- Remove thinking parts from the content list so they don't get encoded as nil

This matches the OpenAI/Moonshot API contract where reasoning_content lives at the message level, not inside the content array.

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -847,7 +847,7 @@ defmodule ReqLLM.Provider.Defaults do
     |> maybe_add_field(:tool_calls, tc)
     |> maybe_add_field(:tool_call_id, tcid)
     |> maybe_add_field(:name, name)
-    |> maybe_add_field(:reasoning_content, reasoning_content)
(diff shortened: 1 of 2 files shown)


mikehostetler:
Thanks for the focused fix here. I marked this `needs_work` because the encoder change needs a small regression test proving assistant thinking parts are emitted as top-level `reasoning_content` and removed from the content array. It would also help to either scope that field to the OpenAI-compatible providers/models that accept it, or add a test/note showing the shared encoder behavior is intentional across those providers. Appreciate the contribution; this looks close once that coverage and scope are tightened up.

PR #660, opened 2026-04-30
fix(deepseek): inject reasoning_content to all assistant messages

BillHuang2001:
## Description

DeepSeek v4 requires the `reasoning_content` field to be passed back in the message history, even if the value is empty. While a previous commit (735bc83) addressed part of this requirement, an edge case remained: 

*   **The Issue:** If an assistant replies without a "thinking" phase, the resulting message object lacks the `reasoning_content` key entirely. During the subsequent turn, the API rejects the request because the `reasoning_content` in missing from the previous assistant message.
*   **The Solution:** This PR ensures that an empty string is explicitly assigned to `reasoning_content` for all assistant messages where the field is currently missing.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests pass (`mix test`)

## Checklist

- [ ] I have added tests that prove my fix/feature works

## Related Issues

Closes #

### DIFF
--- a/lib/req_llm/providers/deepseek.ex
+++ b/lib/req_llm/providers/deepseek.ex
@@ -60,13 +60,7 @@ defmodule ReqLLM.Providers.Deepseek do
           nil
 
         msgs when is_list(msgs) ->
-          Enum.map(msgs, fn msg ->
-            if Map.get(msg, :role) == "assistant" or Map.get(msg, "role") == "assistant" do
-              Map.put_new(msg, :reasoning_content, "")
-            else
-              msg
-            end
-          end)
+          Enum.map(msgs, &ensure_assistant_reasoning_content/1)
 
         other ->
           other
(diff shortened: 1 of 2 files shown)


mikehostetler:
Thanks for putting this together. The DeepSeek follow-up-turn issue is a real edge case, and the approach of normalizing assistant messages in the provider body builder makes sense.

Before this is ready to merge, could you make two small follow-ups?

1. Please remove the inline comments inside `build_body/1`. This repo has a strict no-comments-inside-function-bodies rule, and the logic should be covered by naming/tests instead.
2. Please add a focused regression test for DeepSeek encoding that verifies a prior assistant message without reasoning is sent with `reasoning_content: ""` on the next request.

CI is green and the PR is merge-clean, so once those are addressed this should be straightforward to re-review.

BillHuang2001:
Thanks for the feedback. The two issued have been addressed.

PR #666, opened 2026-05-02
Add MiniMax provider

preciz:
## Description

Add MiniMax provider

## Type of Change

## Testing

- [x] Tests pass (`mix test`)

## Checklist

##

PASSING FIXTURES (6):                                                                                                            
   ✓ MiniMax-M2.7-highspeed                                                                                                       
   ✓ MiniMax-M2.7                                                                                                                 
   ✓ MiniMax-M2.5-highspeed                                                                                                       
   ✓ MiniMax-M2.5                                                                                                                 
   ✓ MiniMax-M2.1                                                                                                                 
   ✓ MiniMax-M2


mikehostetler:
Thanks for the MiniMax provider work, Barna. This is a solid addition and the fixture coverage is helpful.

I found two blocking issues that need investigation before this can merge:

1. MiniMax reasoning details are normalized into `ReqLLM.Message.ReasoningDetails`, but they are not converted back to MiniMax's wire shape when assistant messages are sent again in conversation history. The recorded tool round-trip fixture shows the next request sending fields like `signature`, `provider`, and `provider_data` instead of MiniMax's expected `type`, `id`, `format`, `index`, and `text` fields. Since MiniMax requires full `reasoning_details` preservation for interleaved thinking/tool use, this likely breaks reasoning continuity.

2. Streamed `reasoning_details` fragments are normalized independently per chunk. MiniMax appears to stream progressive fragments for the same reasoning item, but the default response builder appends them as separate reasoning details instead of assembling one final detail. That makes streaming and non-streaming responses diverge and can also produce malformed history if the streamed response is reused.

Could you investigate these and add regression coverage for the MiniMax round-trip/history shape and streamed reasoning assembly?

PR #708, opened 2026-05-19
feat(openai): include encrypted reasoning content

### LINKED ISSUE #693: [Feature]: Send `include: ["reasoning.encrypted_content"]` on OpenAI Responses API requests
### Feature Type

Provider feature extension

### Provider (if applicable)

openai

### Description

PR #344 added round-trip plumbing for reasoning signatures across all three providers. For Anthropic and Google it works as-is because they return signatures inline by default. For OpenAI's Responses API it doesn't, because OpenAI only emits `encrypted_content` on reasoning output items when the request body has `include: ["reasoning.encrypted_content"]`, and ReqLLM never sends that.

I grepped the OpenAI provider source on `main`. `"include"` doesn't appear anywhere in `responses_api.ex` outside of tests. The `@provider_schema` in `openai.ex` (lines 118-238) lists 24 documented options and `include` isn't one. The `build_request_body` whitelist in `responses_api.ex` (lines 660-680) covers `model`, `input`, `stream`, `max_output_tokens`, `reasoning`, `tools`, `tool_choice`, `parallel_tool_calls`, `service_tier`, `text`, `previous_response_id`, `store`. That's the full list.

So in practice on OpenAI, `decode_reasoning_segment` (line 2114) always sees `data["encrypted_content"] == nil`, the resulting message carries `signature: nil`, and on the next turn `encode_reasoning_item_with_signature` (line 841) has nothing to put back. The round-trip code is in place, it just never carries data.

This bites flows that do manual context management on OpenAI reasoning models. ZDR orgs are the obvious case, but anything that builds the `input:` array per turn (instead of relying on `store: true` + `previous_response_id`) is in the same spot. Without the include flag, the reasoning chain doesn't carry across turns. The model re-reasons every turn, tokens go up, quality goes down.

The fix is small. Expose `include` as a provider option so callers can pass it explicitly, and ideally default it to `["reasoning.encrypted_content"]` when the model is a reasoning model so #344's round-trip works out of the box. Keeping it as a list also leaves room for the other include values the Responses API supports (`file_search_call.results`, `web_search_call.results`, `code_interpreter_call.outputs.logs`, etc.).

OpenAI docs: https://platform.openai.com/docs/api-reference/responses/create#responses_create-include

### Example Usage

```elixir
# Explicit caller control
ReqLLM.stream_text(
  "openai:gpt-5-mini",
  context,
  provider_options: [include: ["reasoning.encrypted_content"]]
)

# Or, ideally, automatic for reasoning models so #344 works out of the box
# without any caller-side config.
```

### Can you contribute this?

Yes, with guidance

mikehostetler:
## Summary
- add OpenAI `include` provider option support
- pass `include` through to Responses API request bodies exactly when callers provide it
- default Responses API reasoning models to `include: ["reasoning.encrypted_content"]` when omitted, while preserving explicit empty overrides

Fixes #693

## Validation
- `mix test test/provider/openai/responses_api_unit_test.exs`
- `mix quality`
- `mix test`

### DIFF
--- a/lib/req_llm/providers/openai.ex
+++ b/lib/req_llm/providers/openai.ex
@@ -150,6 +150,11 @@ defmodule ReqLLM.Providers.OpenAI do
       type: :integer,
       doc: "Maximum output tokens for Responses API models"
     ],
+    include: [
+      type: {:list, :string},
+      doc:
+        "Responses API include values, such as reasoning.encrypted_content for reasoning signatures"
+    ],
     openai_structured_output_mode: [
       type: {:in, [:auto, :json_schema, :tool_strict]},
       default: :auto,
--- a/lib/req_llm/providers/openai/responses_api.ex
+++ b/lib/req_llm/providers/openai/responses_api.ex
@@ -75,6 +75,7 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
     "x_search_call" => :x_search_call
   }
   @assistant_phases ["commentary", "final_answer"]
+  @reasoning_encrypted_content_include ["reasoning.encrypted_content"]
 
   @impl true
   def path, do: "/responses"
@@ -680,6 +681,7 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
     tool_choice = encode_tool_choice(opts_map[:tool_choice])
     reasoning = encode_reasoning_effort(opts_map[:reasoning_effort])
     service_tier = opts_map[:service_tier] || provider_opts[:service_tier]
+    include = response_include(provider_opts, model_name)
 
     text_format = encode_text_format(provider_opts[:response_format], provider_opts[:verbosity])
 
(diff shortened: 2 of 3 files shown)


mikehostetler:
Hardening pass found and fixed one streaming gap after the initial PR: `stream_text/3` was sending the default `include`, but `response.completed` reasoning output items were not being surfaced into `StreamResponse.to_response/1` as `reasoning_details`.

Follow-up commit `5d92608` now extracts encrypted reasoning details from completed Responses API stream output and adds regression coverage for that event shape.

Additional validation:
- Live OpenAI non-streaming reasoning call on `openai:gpt-5-mini` returned one signed encrypted reasoning detail.
- Live OpenAI streaming reasoning call on `openai:gpt-5-mini` now returns one signed encrypted reasoning detail through `StreamResponse.to_response/1`.
- `mix test test/provider/openai/responses_api_unit_test.exs`: 111 tests, 0 failures.
- `mix test`: 3186 tests, 0 failures.
- `mix quality`: passed, including Dialyzer.
- GitHub CI is green across the full matrix.

PR #709, opened 2026-05-20
feat: Add spans wrapper to OpenTelemetry

vasspilka:
## Description

Adding spans to group relevant traces together.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests pass (`mix test`)

## Checklist

## Related Issues

Closes #

### DIFF
--- a/lib/req_llm/open_telemetry.ex
+++ b/lib/req_llm/open_telemetry.ex
@@ -179,33 +179,46 @@ defmodule ReqLLM.OpenTelemetry.OTelAdapter do
   end
 
   @impl true
-  def start_child_span(parent, name, attributes, opts, _config) do
-    # Erlang OTel conveys the parent via the Ctx argument of
-    # `:otel_tracer.start_span/4`, not via the start_opts map. We derive a
-    # child Ctx from the current one (preserving baggage / propagators)
-    # and override its current-span slot with the supplied parent. The
-    # span returned by `:otel_tracer.start_span/4` will pick that up as
-    # its parent.
-    ctx = call(:otel_ctx, :get_current, [])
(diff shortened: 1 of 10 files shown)


mikehostetler (changes requested):
This needs work before merge.

Findings:

1. Blocking: CI is red. `mix credo --strict` fails on `lib/req_llm/stream_server.ex:927`; the new `cond` has only one real condition plus `true`, so Credo requires `if` or pattern matching here.

2. Blocking: the new OpenTelemetry child-span path has no regression coverage. This PR changes Responses API stream decoding, stream timing accumulation, OpenTelemetry mapping, and adapter translation, but no tests assert that a builtin `web_search_call`/`file_search_call` becomes a `tool_spans` entry or invokes the adapter child-span callbacks. Please add focused tests around the mapper/translator and, ideally, a stream-chunk timing test.

Validation:
- GitHub CI: Credo failing; other checks passed.
- Local telemetry slice passes: `mix test test/req_llm/telemetry --max-failures 1`.

Leaving this as `needs_work`.

mikehostetler (approved):
Patched and re-reviewed. The previous blockers are resolved: Credo is green, builtin tool timing no longer leaks internal stream markers, and the new child-span path now has focused coverage across Responses API decoding, stream timing, the OpenTelemetry mapper, and the translator. Local validation and GitHub CI are green.

vasspilka:
Thanks @mikehostetler I will do a before/after run with Langfuse and post the screenshots here so the difference is visible

mikehostetler (approved):
Second hardening pass complete. Added a follow-up fix for mixed atom/string builtin timing metadata, omitted empty timing metadata from stream stop events, and guarded child-span timestamps so partial timing does not emit invalid explicit end times. Local full suite and GitHub CI are green.

PR #714, opened 2026-05-21
fix: use actual provider prefix in ensure_deep_research_tools LLMDB lookup

noozo:
## Summary

- `ensure_deep_research_tools/2` in `ResponsesAPI` hardcoded `"openai:"` when looking up a model in LLMDB, even when the request was routed through a different provider (e.g. xAI)
- xAI Live Search routes through `ResponsesAPI.encode_body/1` — so a model like `xai:grok-4-fast` triggered a lookup for `openai:grok-4-fast`, which is not in the LLMDB catalog, emitting a spurious `"Using unverified model"` warning
- Added `provider: "xai"` to xAI request options so the lookup uses the correct catalog entry
- `ensure_deep_research_tools` now uses `request.options[:provider] || "openai"` — OpenAI requests are unaffected

## Root cause

```
xAI.prepare_chat_request → stores model: "grok-4-fast" (bare id, no provider)
xAI.encode_responses_body → ResponsesAPI.encode_body
ResponsesAPI.ensure_deep_research_tools → ReqLLM.model("openai:grok-4-fast")  ← wrong prefix
LLMDB lookup fails → "Using unverified model" warning
```

## Test plan

- [ ] xAI Responses API call with `xai_tools: [%{type: "web_search"}]` no longer emits `"Using unverified model: openai:grok-4-fast"`
- [ ] OpenAI Responses API calls continue to resolve models correctly (fallback to `"openai"` prefix)
- [ ] Deep-research model auto-injection still works for OpenAI deep-research models

### PS: @mikehostetler Unsure this is a proper fix at all, in the sense that it feels wrong to have a provider key inside the xai provider :)

### DIFF
--- a/lib/req_llm/providers/openai/responses_api.ex
+++ b/lib/req_llm/providers/openai/responses_api.ex
@@ -1352,9 +1352,10 @@ defmodule ReqLLM.Providers.OpenAI.ResponsesAPI do
 
   defp ensure_deep_research_tools(tools, request) do
     model_name = request.options[:model] || request.options[:id]
+    provider = request.options[:provider] || "openai"
 
     if is_binary(model_name) and model_name != "" do
-      case ReqLLM.model("openai:#{model_name}") do
+      case ReqLLM.model("#{provider}:#{model_name}") do
         {:ok, model} ->
           category = get_in(model, [Access.key(:extra, %{}), :category])
 
--- a/lib/req_llm/providers/xai.ex
+++ b/lib/req_llm/providers/xai.ex
@@ -282,7 +282,7 @@ defmodule ReqLLM.Providers.XAI do
 
       req_keys =
         supported_provider_options() ++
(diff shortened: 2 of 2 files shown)


mikehostetler:
Pushed a hardened version to agentjido:harden/pr-714-xai-provider-lookup (commit 7ca103e). I could not push to the fork (Use-Passionfruit/req_llm) despite maintainerCanModify=true; GitHub returned permission denied. This branch removes the request :provider option, carries the resolved model internally for non-streaming and streaming, and adds regression coverage. Validation: mix test test/providers/xai_test.exs, mix test test/provider/openai/responses_api_unit_test.exs, mix quality.

mikehostetler:
Closed as superseded by #715, which includes this fix plus the streaming hardening and regression tests. #715 was merged in 900f53a.

PR #715, opened 2026-05-22
fix: preserve xAI model metadata in Responses API lookup

mikehostetler:
## Summary
- carry resolved model metadata into the shared Responses API deep-research tool lookup
- preserve xAI model metadata for both non-streaming and streaming Responses API requests
- add xAI regressions that assert the OpenAI fallback warning does not return

## Validation
- mix test test/providers/xai_test.exs
- mix test test/provider/openai/responses_api_unit_test.exs
- mix quality

Supersedes #714, which could not be pushed to the contributor fork from this environment.


noozo:
<3

PR #737, opened 2026-06-01
feat(anthropic): honor explicit cache_control from ContentPart metadata

ArthV:
## Summary

The Anthropic context encoder discards any `cache_control` set on a `ContentPart`, so prompt caching can only be driven by the `anthropic_prompt_cache` / `anthropic_prompt_cache_ttl` options. Those stamp a **single** TTL on the last block (system / tools / message), which cannot express Anthropic's [explicit cache breakpoints](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#explicit-cache-breakpoints) with **per-block TTLs** — e.g. a 1-hour system block followed by a 5-minute one.

This PR lets callers declare a cache breakpoint directly on a content block via its `metadata`, and the encoder carries it onto the emitted block:

```elixir
ReqLLM.Context.system([
  ContentPart.text(style_guide, %{cache_control: %{type: "ephemeral", ttl: "1h"}}),
  ContentPart.text(context,     %{cache_control: %{type: "ephemeral"}})  # 5m default
])
```

## Changes

- `encode_content_part/1` now applies a `cache_control` found in a ContentPart's `metadata` (atom **or** string key) onto the encoded block, for every block type (text/image/file/document). Implemented as a thin wrapper over the existing per-type encoders (renamed `do_encode_content_part/1`) — the shape logic is unchanged.
- `normalize_system_content/1` only collapses a lone text block to a bare string when it carries nothing else (`map_size == 2`), so an explicit cache breakpoint on a **single** system block is no longer silently stripped.

It composes with the existing auto-injection (`maybe_apply_prompt_caching/2` only adds `cache_control` to a block that lacks one), so there is **no behavior change** for current callers.

## Tests

Added a `describe "explicit per-block cache_control via ContentPart metadata"` block to `anthropic_prompt_cache_test.exs` covering: a single explicit breakpoint with a TTL, multiple breakpoints with distinct TTLs (1h before 5m), and the no-cache_control passthrough. Full suite green (`3340 tests, 0 failures`).

## Verification

Verified end-to-end against `claude-opus-4-6` on Google Vertex: the request is accepted with mixed TTLs, and the two breakpoints produce independent cache entries (the 1-hour block is read from cache while the trailing 5-minute block is rewritten when its content changes).

### DIFF
--- a/lib/req_llm/providers/anthropic/context.ex
+++ b/lib/req_llm/providers/anthropic/context.ex
@@ -217,8 +217,7 @@ defmodule ReqLLM.Providers.Anthropic.Context do
 
     case content_blocks do
       [] -> ""
-      # Simplify single text blocks
-      [%{type: "text", text: text}] -> text
+      [%{type: "text", text: text} = block] when map_size(block) == 2 -> text
       blocks -> blocks
     end
   end
--- a/lib/req_llm/providers/anthropic/context.ex
+++ b/lib/req_llm/providers/anthropic/context.ex
@@ -196,7 +196,13 @@ defmodule ReqLLM.Providers.Anthropic.Context do
   defp blank_system_content_block?(_block), do: false
 
   defp normalize_system_content([]), do: nil
-  defp normalize_system_content([%{type: "text", text: text}]), do: text
+  # Collapse a lone plain text block to a bare string — but only when it carries
+  # nothing else (e.g. no explicit `cache_control`), so an explicit cache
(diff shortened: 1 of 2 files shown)


ArthV:
@mikehostetler it's vicoded as you can see, I'm open to any change that could help achieve the same thing. The main idea here is to be able to control the cache at the message level using Anthropic's models

mikehostetler:
@ArthV all good! always appreciate stuff like this

PR #749, opened 2026-06-04
fix(anthropic): map refusal/pause_turn/model_context_window_exceeded stop_reasons instead of collapsing to :error

ArthV:
## Problem

The Anthropic provider only recognized a subset of `stop_reason` values. Anything else was collapsed to `:error` in the non-streaming decoder:

```elixir
defp parse_finish_reason(reason) when is_binary(reason), do: :error
```

This has two consequences:

1. **Lost signal.** Newer Anthropic stop reasons — notably `"refusal"` (the model declined for safety), plus `"pause_turn"` and `"model_context_window_exceeded"` — were all flattened to `:error`.
2. **Non-errors surfaced as errors.** A `refusal` is a normal (if unfortunate) completion with an empty body, not a transport/decode failure. Reporting it as `finish_reason: :error` led a downstream caller to treat a perfectly valid `%ReqLLM.Response{}` (empty text) as success and then crash on `nil` text. The real reason was invisible until we logged the raw `stop_reason` and saw `"refusal"`.

On top of that, the **streaming** paths (`decode_stream_event/2` and `message_delta_chunks/2`) used their *own* `case` mapping that recognized `stop_sequence` but not `content_filter`/`refusal`/etc., and fell back to `:unknown` — inconsistent with the non-streaming path's `:error`.

## Change

In `lib/req_llm/providers/anthropic/response.ex`:

- Add mappings: `stop_sequence -> :stop`, `model_context_window_exceeded -> :length`, `pause_turn -> :incomplete`, `refusal -> :content_filter`.
- Change the unrecognized-binary fallback from `:error` to `:unknown`, reserving `:error` for genuine decode/transport failures (matches the streaming path's intent).
- Unify the two streaming `message_delta` mappings to reuse `parse_finish_reason/1` (`... || :unknown`) so coverage can't drift between code paths.
- Add tests covering the full `stop_reason` set.

`refusal -> :content_filter` reuses the existing canonical atom (a safety stop) rather than introducing a new `:refusal` atom; happy to add a dedicated atom instead if maintainers prefer.

## Tests

`mix test test/providers/anthropic_test.exs` — 91 passing, including a new case asserting every `stop_reason` maps as expected. `mix format` and `mix credo` clean.

### DIFF
--- a/lib/req_llm/providers/anthropic/response.ex
+++ b/lib/req_llm/providers/anthropic/response.ex
@@ -421,10 +421,6 @@ defmodule ReqLLM.Providers.Anthropic.Response do
   defp parse_finish_reason("pause_turn"), do: :incomplete
   defp parse_finish_reason("refusal"), do: :content_filter
   defp parse_finish_reason("content_filter"), do: :content_filter
-  # An unrecognized but valid stop_reason is "unknown", not an error — reserve
-  # :error for genuine decode/transport failures. (Previously any unmapped
-  # reason — e.g. "refusal", "pause_turn" — collapsed to :error, losing the
-  # signal and surfacing a non-error turn as an error.)
   defp parse_finish_reason(reason) when is_binary(reason), do: :unknown
   defp parse_finish_reason(_), do: nil
 
--- a/lib/req_llm/providers/anthropic/response.ex
+++ b/lib/req_llm/providers/anthropic/response.ex
@@ -100,14 +100,7 @@ defmodule ReqLLM.Providers.Anthropic.Response do
         [ReqLLM.StreamChunk.meta(%{terminal?: true})]
 
       %{"type" => "message_delta", "delta" => delta} ->
-        finish_reason =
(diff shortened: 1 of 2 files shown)


mikehostetler:
Thanks for the clear fix and the detailed stop-reason coverage. The mapping change looks useful, and CI is green.

I marked this `needs_work` because the patch adds comments inside function/test bodies, which this repo explicitly disallows. Please remove those in-body comments or move the explanation into test names/module docs. Once that style issue is cleaned up, this looks close.

ArthV:
Thanks for the guidance. 

Let me know if I can do anything else to get it merged

mikehostetler:
Thanks for the quick follow-up. The added style commit removes the in-body comments I flagged, and the remaining Anthropic stop-reason mapping/test changes look scoped and covered. CI is green and the PR is merge-clean against current `main`, so I marked this `ready_to_merge`.

Issue #728, opened 2026-05-26
[Improvement]: ReqLLM can leave JSON encoding to Req's `:encode` request step


razielgn:
### Feature Type

Core library enhancement

### Description

Two improvements:
- Leave JSON encoding to Req's `encode` step: it uses `Jason.encode_to_iodata`, which is more efficient.
- Other added steps, for example PostHog's https://hexdocs.pm/posthog/PostHog.Integrations.LLMAnalytics.Req.html, must read the soon-to-be-json-encoded body to work properly. The current workaround would be to parse the body that has just been encoded, but it's inefficient.

If I'm not mistaken, the streaming mode goes through Finch directly, so I guess there is no other choice there, but at least that part should also use `Jason.encode_to_iodata`.

Tested:

```diff
diff --git a/lib/req_llm/providers/anthropic.ex b/lib/req_llm/providers/anthropic.ex
index 1b8e19a..7d89be4 100644
--- a/lib/req_llm/providers/anthropic.ex
+++ b/lib/req_llm/providers/anthropic.ex
@@ -365,7 +365,7 @@ defmodule ReqLLM.Providers.Anthropic do
     )
     |> ReqLLM.Step.Error.attach()
     |> ReqLLM.Step.Retry.attach(user_opts)
-    |> Req.Request.append_request_steps(llm_encode_body: &encode_body/1)
+    |> Req.Request.prepend_request_steps(llm_encode_body: &encode_body/1)
     |> Req.Request.append_response_steps(llm_decode_response: &decode_response/1)
     |> ReqLLM.Step.Usage.attach(model)
     |> ReqLLM.Step.Telemetry.attach(model, user_opts)
@@ -396,9 +396,7 @@ defmodule ReqLLM.Providers.Anthropic do
       |> build_request_body(model_name, opts)
       |> shape_subscription_body(request)
 
-    json_body = body |> ReqLLM.Schema.apply_property_ordering() |> Jason.encode!()
-
-    %{request | body: json_body}
+    put_in(request.options[:json], ReqLLM.Schema.apply_property_ordering(body))
   end
 
   @impl ReqLLM.Provider
```

Obviously all tests break because they expect the body to be there. They can be fixed like this, but I guess it should be refactored.

```diff
diff --git a/test/providers/anthropic_test.exs b/test/providers/anthropic_test.exs
index 90ff7ab..ad296a7 100644
--- a/test/providers/anthropic_test.exs
+++ b/test/providers/anthropic_test.exs
[...]

PR #750, opened 2026-06-04
feat: support adaptive-thinking only Claude models on Azure and Bedrock

### LINKED ISSUE #743: [Feature]: Support for Adaptive-Thinking Only Claude Models on Azure and Amazon Bedrock
### Feature Type

New feature support

### Provider (if applicable)

azure, amazon_bedrock

## Summary

I'm not sure if this is a bug or a feature request, so I have labelled it as a feature request for the moment.

#732 and #728 added support for adaptive-thinking only Claude models (Opus 4.7, Opus 4.8, Mythos Preview) to the Anthropic provider.

However, this support doesn't seem to work when using these models on third party providers, such as Azure AI Foundry and Amazon Bedrock

The following issues currently seem to block this from working:

### Detection of reasoning capability and adaptive thinking requirement

Both Azure and Amazon bedrock check to see if the model supports reasoning before translating reasoning parameters into Claude thinking parameters via `PlatformReasoning.add_reasoning_to_additional_fields/2`

`lib/req_llm/providers/azure/anthropic.ex`

(on a sidenote here, I think we could use `ModelHelpers.reasoning_enabled?(model)` instead of manually checking inside capabilities here?)

```elixir
defp maybe_translate_reasoning_params(model, opts) do
    {reasoning_effort, opts} = Keyword.pop(opts, :reasoning_effort)
    {reasoning_budget, opts} = Keyword.pop(opts, :reasoning_token_budget)

    has_reasoning =
      get_in(model, [Access.key(:capabilities), Access.key(:reasoning), Access.key(:enabled)]) ==
        true

    cond do
      has_reasoning && reasoning_budget && is_integer(reasoning_budget) ->
        opts
        |> PlatformReasoning.add_reasoning_to_additional_fields(reasoning_budget)
        |> ensure_min_max_tokens(reasoning_budget)
        |> set_reasoning_temperature(model)

      has_reasoning && reasoning_effort && reasoning_effort != :none ->
        budget = Anthropic.map_reasoning_effort_to_budget(reasoning_effort)

        opts
        |> PlatformReasoning.add_reasoning_to_additional_fields(budget)
        |> ensure_min_max_tokens(budget)
        |> set_reasoning_temperature(model)

      (reasoning_effort || reasoning_budget) && !has_reasoning ->
        Logger.warning(
          "Reasoning parameters ignored: model #{inspect(model.id)} does not support extended thinking"
        )

        opts

      true ->
        opts
    end
  end
```

`lib/req_llm/providers/amazon_bedrock.ex`

```elixir
defp maybe_translate_reasoning_params(model, opts) do
    model_id = model.provider_model_id || model.id

    # Check if this is a Claude model with reasoning capability
    is_claude = String.contains?(model_id, "anthropic.claude")
    has_reasoning = ModelHelpers.reasoning_enabled?(model)

    if is_claude and has_reasoning do
      {reasoning_effort, opts} = Keyword.pop(opts, :reasoning_effort)
      {reasoning_budget, opts} = Keyword.pop(opts, :reasoning_token_budget)

      cond do
        reasoning_budget && is_integer(reasoning_budget) ->
          # Explicit budget_tokens provided
          PlatformReasoning.add_reasoning_to_additional_fields(opts, reasoning_budget)

        reasoning_effort && reasoning_effort != :none ->
          # Map effort to budget using canonical Anthropic mappings
          budget = Anthropic.map_reasoning_effort_to_budget(reasoning_effort)
          PlatformReasoning.add_reasoning_to_additional_fields(opts, budget)

        true ->
          # No reasoning params or :none (disable reasoning)
          opts
      end
    else
      # Not a Claude reasoning model, pass through
      opts
    end
  end
```

Perhaps I am just "holding it wrong", but `ReqLLM.model!/1` seems to return no capabilities for these models, so im not sure that they would ever be determined to be reasoning capable.

Also, the solution in #732 relies on the model extra metadata to determine that _only_ adaptive thinking is available via `adaptive_thinking_required?/1` in the main Anthropic provider module, so if this data is not available, we can never determine to turn on adaptive thinking.

Returns capabilities and extra data needed to determine adaptive thinking is required:
```elixir
ReqLLM.model!("anthropic:claude-opus-4-8")
```

Does not return capabilities or extra data needed to determine adaptive thinking is required:
```elixir
ReqLLM.model!("azure:claude-opus-4-8")
ReqLLM.model!("amazon_bedrock:claude-opus-4-8")
```

### PlatformReasoning.add_reasoning_to_additional_fields/2

Even if the previous section were to properly detect adaptive thinking being required, both Azure and Amazon Bedrock delegate to `PlatformReasoning.add_reasoning_to_additional_fields/2` which only ever sets `%{type: "enabled", budget_tokens: budget_tokens}`

```elixir
def add_reasoning_to_additional_fields(opts, budget_tokens) do
    # Get existing additional_model_request_fields from provider_options (if any)
    provider_opts = Keyword.get(opts, :provider_options, [])

    additional_fields =
      Keyword.get(provider_opts, :additional_model_request_fields, %{})
      |> Map.put(:thinking, %{type: "enabled", budget_tokens: budget_tokens})

    # Put it back into provider_options
    updated_provider_opts =
      Keyword.put(provider_opts, :additional_model_request_fields, additional_fields)

    Keyword.put(opts, :provider_options, updated_provider_opts)
  end
```

### AdapterHelpers.maybe_add_thinking/2

Even if `%{type: "adaptive"}` were to make it into `additional_model_request_fields` (such as manually putting it there rather than using the reasoning_effort and reasoning_budget_tokens options), I believe it is then later stripped back out by `AdapterHelpers.maybe_add_thinking/2` which actually adds the thinking config to the request, as it explicitly pattern matches on `%{type: "enabled", budget_tokens: budget}`:

```elixir
def maybe_add_thinking(body, opts) do
    # Check if additional_model_request_fields has thinking config
    thinking_config =
      get_in(opts, [:provider_options, :additional_model_request_fields, :thinking])

    tool_choice = opts[:tool_choice]

    # Extended thinking doesn't work when tool_choice forces a specific tool
    forced_tool_choice? =
      case tool_choice do
        %{type: "tool", name: _} -> true
        %{"type" => "tool", "name" => _} -> true
        _ -> false
      end

    case thinking_config do
      %{type: "enabled", budget_tokens: budget} when not forced_tool_choice? ->
        Map.put(body, :thinking, %{type: "enabled", budget_tokens: budget})

      _ ->
        body
    end
  end
```

### Example Usage

_No response_

### Can you contribute this?

Yes, with guidance

dl-alexandre:
## Summary
Support "adaptive-thinking only" Claude models (e.g. Opus 4.7, Opus 4.8, Mythos Preview) when used via Azure and Amazon Bedrock (and Google Vertex) providers.

Previously, the adaptive thinking support (type: "adaptive" + special output effort handling, added for direct Anthropic in #728/#732) only worked for the native `anthropic:` provider. Third-party platforms would fall back to `type: "enabled"`, causing the models to not behave correctly (or error on detection).

### Changes
- Added public `ReqLLM.ModelHelpers.adaptive_thinking_required?/1` (with tests) that checks the model catalog's extra capabilities for `thinking.types.adaptive.supported == true && enabled.supported == false`.
- Updated `ReqLLM.Providers.Anthropic.PlatformReasoning.add_reasoning_to_additional_fields/3` (now accepts optional model) to emit `%{type: "adaptive", display: "summarized"}` instead of the "enabled" + budget form when required.
- Extended `maybe_translate_reasoning_params` (and the no-param fallback) in:
  - `amazon_bedrock`
  - `azure/anthropic`
  - `google_vertex/anthropic`
  to detect adaptive models and use the correct platform config (via the shared PlatformReasoning).
- Minor doc updates.

The change is backward compatible and only activates for models that the catalog marks as adaptive-only.

## Validation
- `mix test test/req_llm/model_helpers_test.exs` (new tests pass)
- Full compile of req_llm
- Manual review of translation paths for all three platforms

Fixes #743

### DIFF
--- a/lib/req_llm/providers/amazon_bedrock.ex
+++ b/lib/req_llm/providers/amazon_bedrock.ex
@@ -728,9 +728,10 @@ defmodule ReqLLM.Providers.AmazonBedrock do
   defp maybe_translate_reasoning_params(model, opts) do
     model_id = model.provider_model_id || model.id
 
-    # Check if this is a Claude model with reasoning capability
     is_claude = String.contains?(model_id, "anthropic.claude")
-    has_reasoning = ModelHelpers.reasoning_enabled?(model)
+
+    has_reasoning =
+      ModelHelpers.reasoning_enabled?(model) or ModelHelpers.adaptive_thinking_required?(model)
 
     if is_claude and has_reasoning do
       {reasoning_effort, opts} = Keyword.pop(opts, :reasoning_effort)
@@ -738,25 +739,16 @@ defmodule ReqLLM.Providers.AmazonBedrock do
 
       cond do
         reasoning_budget && is_integer(reasoning_budget) ->
-          # Explicit budget_tokens provided
           PlatformReasoning.add_reasoning_to_additional_fields(opts, reasoning_budget, model)
 
         reasoning_effort && reasoning_effort != :none ->
-          # Map effort to budget using canonical Anthropic mappings
           budget = Anthropic.map_reasoning_effort_to_budget(reasoning_effort)
           PlatformReasoning.add_reasoning_to_additional_fields(opts, budget, model)
 
         true ->
-          # No reasoning params or :none (disable reasoning)
-          # Still check if model requires adaptive thinking (for "thinking-only" models)
-          if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
-            PlatformReasoning.add_reasoning_to_additional_fields(opts, nil, model)
-          else
-            opts
-          end
+          opts
       end
     else
-      # Not a Claude reasoning model, pass through
       opts
     end
   end
--- a/lib/req_llm/providers/amazon_bedrock.ex
+++ b/lib/req_llm/providers/amazon_bedrock.ex
@@ -739,16 +739,21 @@ defmodule ReqLLM.Providers.AmazonBedrock do
       cond do
         reasoning_budget && is_integer(reasoning_budget) ->
           # Explicit budget_tokens provided
-          PlatformReasoning.add_reasoning_to_additional_fields(opts, reasoning_budget)
+          PlatformReasoning.add_reasoning_to_additional_fields(opts, reasoning_budget, model)
 
         reasoning_effort && reasoning_effort != :none ->
           # Map effort to budget using canonical Anthropic mappings
           budget = Anthropic.map_reasoning_effort_to_budget(reasoning_effort)
-          PlatformReasoning.add_reasoning_to_additional_fields(opts, budget)
+          PlatformReasoning.add_reasoning_to_additional_fields(opts, budget, model)
 
         true ->
           # No reasoning params or :none (disable reasoning)
-          opts
+          # Still check if model requires adaptive thinking (for "thinking-only" models)
+          if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
+            PlatformReasoning.add_reasoning_to_additional_fields(opts, nil, model)
+          else
+            opts
+          end
       end
     else
       # Not a Claude reasoning model, pass through
--- a/lib/req_llm/providers/anthropic.ex
+++ b/lib/req_llm/providers/anthropic.ex
@@ -1375,7 +1375,7 @@ defmodule ReqLLM.Providers.Anthropic do
   end
 
   defp put_reasoning_effort(opts, model, effort, reasoning_budget) do
-    if adaptive_thinking_required?(model) do
+    if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
       opts
       |> Keyword.put(:thinking, %{type: "adaptive", display: "summarized"})
       |> put_output_effort(adaptive_effort(effort, model))
@@ -1391,7 +1391,7 @@ defmodule ReqLLM.Providers.Anthropic do
   end
 
   defp put_default_reasoning_effort(opts, model) do
-    if adaptive_thinking_required?(model) do
+    if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
       opts
       |> Keyword.put(:thinking, %{type: "adaptive", display: "summarized"})
       |> put_output_effort(adaptive_effort(:default, model))
@@ -1404,7 +1404,7 @@ defmodule ReqLLM.Providers.Anthropic do
   end
 
   defp normalize_thinking_for_model(opts, model) do
-    if adaptive_thinking_required?(model) do
+    if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
       case Keyword.get(opts, :thinking) do
         %{type: "enabled"} = thinking ->
           opts
@@ -1478,10 +1478,6 @@ defmodule ReqLLM.Providers.Anthropic do
     else
       "high"
     end
-  end
-
-  defp adaptive_thinking_required?(model) do
-    ReqLLM.ModelHelpers.adaptive_thinking_required?(model)
   end
 
   defp remove_model_unsupported_parameters(opts, model) do
--- a/lib/req_llm/providers/anthropic.ex
+++ b/lib/req_llm/providers/anthropic.ex
@@ -1481,8 +1481,7 @@ defmodule ReqLLM.Providers.Anthropic do
   end
 
   defp adaptive_thinking_required?(model) do
-    model_extra(model, [:capabilities, :thinking, :types, :adaptive, :supported]) == true and
-      model_extra(model, [:capabilities, :thinking, :types, :enabled, :supported]) == false
+    ReqLLM.ModelHelpers.adaptive_thinking_required?(model)
   end
 
   defp remove_model_unsupported_parameters(opts, model) do
--- a/lib/req_llm/providers/azure/anthropic.ex
+++ b/lib/req_llm/providers/azure/anthropic.ex
@@ -53,6 +53,7 @@ defmodule ReqLLM.Providers.Azure.Anthropic do
   alias ReqLLM.Providers.Anthropic
   alias ReqLLM.Providers.Anthropic.AdapterHelpers
   alias ReqLLM.Providers.Anthropic.PlatformReasoning
+  alias ReqLLM.ModelHelpers
 
   require Logger
 
@@ -298,9 +299,7 @@ defmodule ReqLLM.Providers.Azure.Anthropic do
     {reasoning_effort, opts} = Keyword.pop(opts, :reasoning_effort)
     {reasoning_budget, opts} = Keyword.pop(opts, :reasoning_token_budget)
 
-    has_reasoning =
-      get_in(model, [Access.key(:capabilities), Access.key(:reasoning), Access.key(:enabled)]) ==
-        true
+    has_reasoning = ModelHelpers.reasoning_enabled?(model)
 
     cond do
       has_reasoning && reasoning_budget && is_integer(reasoning_budget) ->
--- a/lib/req_llm/providers/azure/anthropic.ex
+++ b/lib/req_llm/providers/azure/anthropic.ex
@@ -50,10 +50,10 @@ defmodule ReqLLM.Providers.Azure.Anthropic do
   Usage information is automatically included in streaming responses.
   """
 
+  alias ReqLLM.ModelHelpers
   alias ReqLLM.Providers.Anthropic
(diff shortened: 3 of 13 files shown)


lib/req_llm/providers/azure/anthropic.ex
green-david:
```suggestion
has_reasoning = ReqLLM.ModelHelpers.reasoning_enabled?(model)
```

lib/req_llm/providers/azure/anthropic.ex
```diff
@@ -325,7 +325,14 @@ defmodule ReqLLM.Providers.Azure.Anthropic do
...
       true ->
-        opts
+        # Check for adaptive thinking only models even without explicit params
+        if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
+          opts
+          |> PlatformReasoning.add_reasoning_to_additional_fields(nil, model)
+          |> set_reasoning_temperature(model)
+        else
+          opts
+        end
```
green-david: Hmm, im not so sure about this one. I think the only model that _has_ to think is Mythos Preview, which defaults to adaptive thinking and won't allow thinking to be disabled.

Wouldn't this mean even if no reasoning parameters were passed, we would turn on adaptive thinking for the user for Opus 4.7 and Opus 4.8?

green-david: I think this is perhaps due to a misnomer of `adaptive_thinking_required?`, when it is really not _required_ to perform adaptive thinking, it is just that only adaptive thinking is supported if you _do_ want it to think.

lib/req_llm/providers/google_vertex/anthropic.ex
```diff
@@ -202,13 +202,20 @@ defmodule ReqLLM.Providers.GoogleVertex.Anthropic do
...
           # No reasoning params or :none (disable reasoning)
-          opts
+          # Still support adaptive for "thinking-only" models
+          if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
+            opts
+            |> PlatformReasoning.add_reasoning_to_additional_fields(nil, model)
+            |> Keyword.put(:temperature, 1.0)
+          else
+            opts
+          end
```
green-david: Same concern here regarding turning on adaptive thinking when the user didn't ask for it

lib/req_llm/providers/amazon_bedrock.ex
```diff
@@ -739,16 +739,21 @@ defmodule ReqLLM.Providers.AmazonBedrock do
...
 
         true ->
           # No reasoning params or :none (disable reasoning)
-          opts
+          # Still check if model requires adaptive thinking (for "thinking-only" models)
+          if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
+            PlatformReasoning.add_reasoning_to_additional_fields(opts, nil, model)
+          else
+            opts
+          end
```
green-david: Same concern here re: turning it on for them

lib/req_llm/providers/amazon_bedrock.ex
```diff
@@ -739,16 +739,21 @@ defmodule ReqLLM.Providers.AmazonBedrock do
...
           budget = Anthropic.map_reasoning_effort_to_budget(reasoning_effort)
-          PlatformReasoning.add_reasoning_to_additional_fields(opts, budget)
+          PlatformReasoning.add_reasoning_to_additional_fields(opts, budget, model)
 
         true ->
           # No reasoning params or :none (disable reasoning)
-          opts
+          # Still check if model requires adaptive thinking (for "thinking-only" models)
+          if ReqLLM.ModelHelpers.adaptive_thinking_required?(model) do
+            PlatformReasoning.add_reasoning_to_additional_fields(opts, nil, model)
```
green-david: this one doesn't need to set the temperature?

lib/req_llm/providers/anthropic.ex
```diff
@@ -1481,8 +1481,7 @@ defmodule ReqLLM.Providers.Anthropic do
   end
 
   defp adaptive_thinking_required?(model) do
-    model_extra(model, [:capabilities, :thinking, :types, :adaptive, :supported]) == true and
-      model_extra(model, [:capabilities, :thinking, :types, :enabled, :supported]) == false
+    ReqLLM.ModelHelpers.adaptive_thinking_required?(model)
   end
```
green-david: Can we eliminate this private function and inline the ModelHelpers call?

dl-alexandre:
@green-david @chatgpt-codex-connector Thanks for the quick reviews!

**Changes pushed:**
- `adaptive_thinking_required?/1` now strictly checks `enabled.supported == false` (explicit). This prevents treating models with omitted `enabled` key in sparse catalog metadata as adaptive-only (addresses codex P2 and your concern for Opus 4.7/4.8). Only models explicitly marked enabled:false (e.g. Mythos Preview that must use adaptive) will match and auto-inject in the no-param fallback.
- Renamed? kept the fn name for this update (behavior clarified in docs); happy to rename to `adaptive_thinking_only?/1` in follow-up if preferred.
- Azure: switched the `has_reasoning` check to `ModelHelpers.reasoning_enabled?(model)` as suggested.
- Direct Anthropic: inlined/eliminated the thin private `adaptive_thinking_required?` wrapper.
- Added regression test for "adaptive true + enabled omitted" -> false.

**On auto-enabling for Opus 4.7/4.8:** with the tightened predicate they should no longer match, so no unexpected adaptive when user passes no reasoning params. For Mythos (the one that "has to think"), the catalog entry should have the explicit false so it continues to work.

**Bedrock temperature:** that path relies on `thinking_constraints` (required 1.0) + shared `maybe_clean_thinking_after_translation`; no manual set in the adaptive no-param branch (unlike Azure which does it locally). Let me know if it needs explicit set there too.

**Other threads:** the suggestion on azure line and inlining done. The "same concern" ones should be mitigated by the predicate fix.

Re-ran model helpers tests (all green). PTAL and resolve threads if happy — thanks!

dl-alexandre:
Addressed feedback from green-david and codex P2s on this PR:
- Tightened + documented adaptive_thinking_required? for explicit false (and atom/string key tolerance).
- Switched to ModelHelpers.reasoning_enabled? helper.
- Inlined private wrapper.
- Removed in-body comments per AGENTS.md.
- Added regression test for sparse metadata case.
Latest commit on branch has the updates. Threads for the addressed points can be resolved; ready for re-review.

dl-alexandre:
Addressed feedback from green-david and codex P2s on this PR:
- Tightened + documented adaptive_thinking_required? for explicit false (and atom/string key tolerance).
- Switched to ModelHelpers.reasoning_enabled? helper.
- Inlined private wrapper.
- Removed in-body comments per AGENTS.md.
- Added regression test for sparse metadata case.
Latest commit on branch has the updates. Threads for the addressed points can be resolved; ready for re-review.

mikehostetler:
Thanks for pushing this adaptive-thinking support across the platform providers. The direction is useful, but I found a blocker and marked the PR `needs_work`.

The new detection relies on nested `model.extra["capabilities"]["thinking"]["types"]` metadata, but the current Bedrock and Google Vertex Claude Opus 4.8 catalog entries I checked only carry basic `extra` fields. That means `adaptive_thinking_required?/1` returns false for those platform models, so the Bedrock/Vertex adaptive fallback will not activate for the models this PR is meant to support. Please either make the detection work with the actual platform metadata or ensure the resolved platform models receive the adaptive metadata, then add request-level tests for the Bedrock/Vertex paths.

mikehostetler:
Thanks for taking on the hosted Claude adaptive-thinking support. I pushed a follow-up commit (`27266640`) that fixes the remaining blocker: adaptive-only detection now handles atom/string metadata and derives sparse Bedrock/Vertex/Azure hosted Claude IDs from the native Anthropic catalog; platform providers only emit adaptive thinking when reasoning was explicitly requested, so no-parameter calls do not turn thinking on.

I also extended the adapter path to carry `type: "adaptive"` thinking into Azure/Vertex request bodies and added request-level coverage for Bedrock, Azure, and Vertex plus helper coverage for sparse hosted models. Validated locally with `mix test test/req_llm/model_helpers_test.exs test/providers/anthropic_test.exs test/req_llm/provider/options_test.exs test/req_llm/providers/amazon_bedrock_test.exs test/provider/azure/azure_test.exs test/providers/google_vertex_gemini_test.exs` and `mix quality`; GitHub CI is green and merge state is clean, so I marked this `ready_to_merge`.

PR #753, opened 2026-06-04
fix: Fix reasoning_effort not being passed through to OpenAI Chat Completion API

green-david:
## Description

This is in the same vein as #244. While testing Groq, I noticed that `reasoning_effort` was not getting passed through and sent to the Chat Completions API for GPT OSS models. The ChatAPI module was retrieving `reasoning_effort` from provider options, but it is a global common option that appears to be handled almost exclusively as a top-level option, not a provider option.

## Type of Change

- [ ] Documentation update

## Breaking Changes

I don't _think_ this is a breaking change as it didn't appear to work as documented in the first place.

## Testing

- [x] Tests pass (`mix test`)

## Checklist

## Related Issues

#244

### DIFF
--- a/guides/groq.md
+++ b/guides/groq.md
@@ -28,7 +28,7 @@ Passed via `:provider_options` keyword:
 - **Type**: `"none"` | `"default"` | `"low"` | `"medium"` | `"high"`
 - **Purpose**: Control reasoning level for compatible models
 - **Compatible**: DeepSeek R1 distill models
-- **Example**: `provider_options: [reasoning_effort: "high"]`
+- **Example**: `reasoning_effort: "high"`
 
 ### `reasoning_format`
 - **Type**: String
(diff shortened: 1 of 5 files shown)


mikehostetler:
Thanks for the contribution here. The implementation is directionally right, the PR is merge-clean, and CI is green.

I marked this `needs_work` because the exact bug fix needs a focused regression test: top-level `reasoning_effort` should make it into the OpenAI Chat Completions request body, not only when nested under `provider_options`. A small provider/unit test around `ChatAPI.encode_body/1` or `OpenAI.prepare_request/4` for a chat-compatible model should be enough. Once that coverage is in, this should be in good shape.

green-david:
@mikehostetler Thanks for taking a look. I force-pushed an update adding a test in `test/providers/openai_test.exs`. Let me know if this needs anything else!

PR #754, opened 2026-06-05
feat: add thinking and reasoning options to deepseek

BillHuang2001:
## Description

Add thinking and reasoning options to deepseek:
* Allow enabling or disabling the thinking
* Allow tweaking the reasoning effort

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests pass (`mix test`)

## Checklist

## Related Issues

Closes #

### DIFF
--- a/lib/req_llm/providers/deepseek.ex
+++ b/lib/req_llm/providers/deepseek.ex
@@ -31,12 +31,59 @@ defmodule ReqLLM.Providers.Deepseek do
       ReqLLM.stream_text("deepseek:deepseek-chat", "Tell me a story")
       |> Enum.each(&IO.write/1)
 
+      # With thinking mode enabled (default for reasoning models)
+      ReqLLM.generate_text("deepseek:deepseek-v4-pro", "Solve this complex problem",
+        provider_options: [
+          thinking: %{type: "enabled"},
+          reasoning_effort: :high
+        ]
+      )
+
+      # With maximum reasoning effort for complex tasks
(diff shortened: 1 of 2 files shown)


mikehostetler:
Thanks for adding DeepSeek reasoning support and the focused provider tests. I marked this `needs_work` because the public request path currently rejects one of the new documented/tested options.

The module docs and direct `translate_options/3` tests cover `reasoning_effort: :max`, but normal request preparation rejects it before translation runs: nested `provider_options: [reasoning_effort: :max]` fails because the DeepSeek provider schema only allows `:thinking`, and top-level `reasoning_effort: :max` fails the core schema because `:max` is not a valid canonical value there. I confirmed this with `Deepseek.prepare_request/4`; the existing tests pass because they call `translate_options/3` and `build_body/1` directly.

Please either make the intended public API accept this path, or document/use the existing canonical `:xhigh` value for max effort, and add a regression around `prepare_request`/`encode_body` so the option is verified through the real user-facing flow.

BillHuang2001:
Thanks for the feedback, i removed the :max options, and everything else is untouched. The :xhigh value is now the max effort level.

mikehostetler:
I pushed a small follow-up commit (`eb8da9fb`) to finish this off. It moves `reasoning_effort` out of `provider_options` in the docs, uses the supported top-level `reasoning_effort: :xhigh` form for max effort, and removes the direct `:max` translation assertion that bypassed normal validation.

Validated locally with `mix test test/providers/deepseek_test.exs`, direct request-path probes for the documented forms, and `mix quality`. GitHub CI is green now, so I marked this `ready_to_merge`.

PR #755, opened 2026-06-05
feat(google-vertex): grammar-constrained structured output for Claude via output_config.format

ArthV:
Generating `:object` output for Claude on Vertex uses a best-effort `structured_output` tool call, which intermittently double-encodes nested arrays as JSON strings (`"options": "[\"a\",\"b\"]"`) — see anthropics/claude-agent-sdk-python#510. Downstream schema validation then rejects the payload.

Add an opt-in `anthropic_structured_output_mode: :json_schema` for the Vertex Anthropic adapter. On `:object`, it sends `output_config: %{format: %{type: "json_schema", schema: <reduced>}}` in the body (instead of the tool) and parses the object from the response message text. Live testing: arrays come back as real arrays every time, vs intermittent stringification before.

Grammar-constrained decoding only accepts a subset of JSON Schema, so the schema is reduced first (`strict_json_schema/1`): strip `min*`/`max*`/`*Items` and the Gemini `propertyOrdering` annotation, force every property required with `additionalProperties: false`. Array length cannot be expressed, so callers must re-validate the count on the response.

Default (`:auto`) is unchanged. `:json_schema` requires the GCP `structured_outputs` partner-model feature to be allow-listed (`constraints/vertexai.allowedPartnerModelFeatures`).

## Description

Brief description of changes.

## Type of Change

- [ ] Documentation update

## Testing

- [ ] Tests pass (`mix test`)

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have added tests that prove my fix/feature works
- [ ] All new and existing tests pass
- [ ] My commits follow conventional commit format

## Related Issues

Closes #

### DIFF
--- a/lib/req_llm/providers/anthropic/adapter_helpers.ex
+++ b/lib/req_llm/providers/anthropic/adapter_helpers.ex
@@ -13,15 +13,18 @@ defmodule ReqLLM.Providers.Anthropic.AdapterHelpers do
   def maybe_add_param(body, key, value), do: Map.put(body, key, value)
 
   @doc """
-  Prepare context and options for :object operations using structured output.
+  Prepare context and options for `:object` operations using a synthetic
+  `structured_output` tool, forcing tool choice to use it.
 
-  Creates a synthetic "structured_output" tool and forces tool choice to use it.
-  This leverages Claude's tool-calling for structured JSON output.
+  This is the best-effort tool-calling path. For grammar-constrained output,
+  callers pass `anthropic_structured_output_mode: :json_schema`, which the
+  platform adapter (e.g. Google Vertex) handles via the response-level
+  `output_config.format` path instead of this tool — see
+  `structured_output_mode/1` and `strict_json_schema/1`.
   """
   def prepare_structured_output_context(context, opts) do
     compiled_schema = Keyword.fetch!(opts, :compiled_schema)
 
-    # Create the structured_output tool
     structured_output_tool =
       ReqLLM.Tool.new!(
         name: "structured_output",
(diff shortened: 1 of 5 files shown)


mikehostetler:
Thanks for pushing the Vertex Claude structured-output path forward. I marked this `needs_work` because I found one functional blocker and one repo-style blocker.

The functional issue is in response decoding: `GoogleVertex.decode_response/1` rebuilds opts with only `:operation` and `:context`, so the new `anthropic_structured_output_mode: :json_schema` setting is not available when `GoogleVertex.Anthropic.parse_response/3` decides whether to parse JSON from response text. In the actual decode path, `AdapterHelpers.structured_output_mode(opts)` falls back to `:auto`, so a valid JSON-schema response leaves `response.object` as `nil`. Please forward the mode/provider options into response parsing and add a regression that exercises `decode_response/1`, not only `format_request/3`.

Also, the patch adds/updates comments inside function bodies in `lib/req_llm/providers/google_vertex/anthropic.ex`, which violates this repo’s AGENTS.md rule. Please move that explanation into docs or make the branches self-describing without in-body comments.

ArthV:
Thanks for the careful review — both addressed in `e78078cc`.

**1. Functional (mode lost in decode):** you're exactly right. `GoogleVertex.decode_response/1` rebuilt the formatter opts with only `:operation` and `:context`, so `anthropic_structured_output_mode` never reached `parse_response/3` → it fell back to `:auto` → tried the tool-call path on a text response → `response.object` was left `nil`. (It was only masked because `Response.unwrap_object/1` has a text-content fallback.)

Fix: `decode_response/1` now forwards `:provider_options` and `:anthropic_structured_output_mode` alongside `:operation`/`:context`, so `:json_schema` responses are parsed from message text and `response.object` is set directly. Added a `decode_response/1` regression test with two cases — mode forwarded → object parsed from text; mode absent → `nil` (guards against the exact regression).

**2. Repo style (in-body comments):** removed the comments I added/updated in `google_vertex/anthropic.ex` and moved the per-branch explanation into the `format_request/3` and `parse_response/3` `@doc`s instead. `mix credo --strict` is clean on the touched files.

mikehostetler:
Thanks for the quick fix. I re-reviewed `e78078cc`; the structured-output mode is now forwarded through `decode_response/1`, the new decode regression covers the exact failure path, and the comments added in the previous patch were moved out of function bodies. GitHub checks are green, merge state is clean, and the targeted tests pass locally, so I marked this `ready_to_merge`.

PR #765, opened 2026-06-11
fix(context): improve error handling around tool calls

razielgn:
## Description

Previously, any `ReqLLM.Error.*` would hit `to_string(error)` and crash with `Protocol.UndefinedError: protocol String.Chars not implemented`

## Type of Change

- [ ] Documentation update

## Breaking Changes

No.

## Testing

- [x] Tests pass (`mix test`)

## Checklist

### DIFF
--- a/lib/req_llm/context.ex
+++ b/lib/req_llm/context.ex
@@ -509,7 +509,15 @@ defmodule ReqLLM.Context do
           append(ctx, tool_result_msg)
 
         {:error, error} ->
-          error_result = %{error: to_string(error)}
+          error_result = %{
+            error:
+              if is_exception(error) do
+                Exception.message(error)
+              else
+                to_string(error)
+              end
+          }
+
           tool_result_msg = tool_result_message(name, id, error_result, %{is_error: true})
           append(ctx, tool_result_msg)
       end
--- a/lib/req_llm/context.ex
+++ b/lib/req_llm/context.ex
@@ -509,19 +509,21 @@ defmodule ReqLLM.Context do
           append(ctx, tool_result_msg)
 
         {:error, error} ->
-          error_result = %{
-            error:
-              if is_exception(error) do
-                Exception.message(error)
-              else
-                to_string(error)
-              end
-          }
+          error_result = %{error: tool_error_message(error)}
 
           tool_result_msg = tool_result_message(name, id, error_result, %{is_error: true})
           append(ctx, tool_result_msg)
       end
     end)
+  end
+
+  defp tool_error_message(error) when is_exception(error), do: Exception.message(error)
+
+  defp tool_error_message(error) do
+    case String.Chars.impl_for(error) do
+      nil -> inspect(error)
+      _impl -> to_string(error)
+    end
   end
 
   defp extract_tool_call_info(%ReqLLM.ToolCall{id: id, function: %{name: name}}), do: {name, id}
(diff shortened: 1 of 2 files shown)


lib/req_llm/context.ex
```diff
@@ -509,7 +509,15 @@ defmodule ReqLLM.Context do
           append(ctx, tool_result_msg)
 
         {:error, error} ->
-          error_result = %{error: to_string(error)}
+          error_result = %{
+            error:
+              if is_exception(error) do
+                Exception.message(error)
+              else
+                to_string(error)
```
mikehostetler: `ReqLLM.Tool.execute/2` allows callbacks to return `{:error, term()}` and passes those values through unchanged. This fallback still crashes for non-exception terms such as `{:error, %{reason: "nope"}}` because maps do not implement `String.Chars`. Use a term-safe formatter here, for example `inspect/1` for non-exception values, and add a regression test for a callback returning a map/tuple error.

mikehostetler (changes requested):
Thanks for tightening the Splode exception path. I found one remaining crash in the same `{:error, error}` branch: callback errors are allowed to be arbitrary terms, and non-exception terms like maps still hit `to_string/1` and raise `Protocol.UndefinedError`. This should be handled before merge.

lib/req_llm/context.ex
razielgn: Ah I see. Is it acceptable to go through `inspect/1`? IIRC it truncates the output after some chars.
It supports `limit: :infinity` but I'd find it scary given that it might accidentally consume a lot of tokens.

mikehostetler:
I prepared the follow-up fix for the remaining callback-error crash, but this PR has `maintainerCanModify: false`, so I can’t push directly to the fork branch.

Fix branch: https://github.com/agentjido/req_llm/tree/pr-765-tool-error-formatting
Commit: https://github.com/agentjido/req_llm/commit/10064f1b

It changes the fallback formatter to handle arbitrary non-exception error terms safely and adds regression coverage for callbacks returning map and tuple errors.

To pull it into this PR branch:

```bash
git fetch https://github.com/agentjido/req_llm.git pr-765-tool-error-formatting
git cherry-pick 10064f1b
```

razielgn:
Done, thanks.

mikehostetler (approved):
Re-checked after the follow-up commit. The callback error formatter now handles exception and non-exception terms safely, the map/tuple regression test covers the previous crash, focused tests pass locally, CI is green, and the branch is merge-clean. The default  fallback is fine here; its defaults are intentionally bounded for collections/printable output, so it avoids the unbounded-token concern while still giving a useful model-visible error.

mikehostetler:
Yes, inspect/1 is acceptable here. The default inspect options are bounded, so we avoid unbounded model-visible output while still preserving a useful representation for maps/tuples and other non-String.Chars terms. I would avoid limit: :infinity for exactly the token-growth reason you called out.

mikehostetler (approved):
Re-checked after the follow-up commit. The callback error formatter now handles exception and non-exception terms safely, the map/tuple regression test covers the previous crash, focused tests pass locally, CI is green, and the branch is merge-clean. The default inspect/1 fallback is fine here; its defaults are intentionally bounded for collections and printable output, so it avoids the unbounded-token concern while still giving a useful model-visible error.

PR #767, opened 2026-06-11
fix(streaming): don't finalize OpenAI-compatible streams at finish_reason

BinaryMuse:
For OpenAI-compatible providers, streaming requests set `stream_options: {include_usage: true}` automatically, and per the OpenAI streaming protocol the final usage payload arrives in a standalone SSE event (`"choices": []`) **after** the event carrying `finish_reason`.

`default_decode_stream_event/2` marked the `finish_reason` chunk `terminal?: true`, and `StreamServer` finalizes the stream as soon as a terminal chunk appears in a batch; the metadata snapshot is frozen and metadata waiters are replied. Whether usage survived therefore depended on network buffering:

- `finish_reason` and the usage event land in the **same HTTP chunk** -> the batch decodes together, usage is captured
- the upstream flushes them in **separate chunks** -> the stream finalizes at `finish_reason`, and the usage event is silently dropped: `StreamResponse.usage/1` returns `nil` and `on_meta` callbacks never see it.

I hit this consistently via OpenRouter when requests were routed to an upstream that flushes the two events separately (and intermittently with other routings); affected requests reported no usage at all, with no error.

The fix removes the `terminal?` mark from the `finish_reason` chunk. It is not the end of an OpenAI-compatible stream; the protocol terminators remain in place and all still finalize:

- the standalone usage chunk (`"choices": []`), already marked terminal
- the `data: [DONE]` event (`termination_event?/1`)
- HTTP stream end (`process_http_event(:done, ...)`)

Worst case, finalization is deferred by one or two SSE events on streams that previously finalized at `finish_reason`.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None. `finish_reason` is still captured into metadata exactly as before; only the early-finalization side effect is removed.

## Testing

- [x] Tests pass (`mix test` — 3412 tests, 0 failures)

New coverage:

- Decode-level test asserting the `finish_reason` chunk is not terminal while the standalone usage chunk still is (`test/req_llm/provider/defaults_test.exs`)
- `StreamServer` end-to-end regression mirroring production: a metadata waiter registered mid-stream, with `finish_reason` and usage delivered as separate `http_event` data chunks (`test/req_llm/stream_server/metadata_test.exs`). Without the fix it fails with the production symptom (`metadata.usage` missing); with it, usage is captured.
- One existing assertion that pinned the old terminal flag updated to the new behavior.

## Checklist

- [x] I have updated the documentation accordingly (inline comment documenting why `finish_reason` is not terminal)

## Related Issues

None filed; happy to open one first if preferred.

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -1150,15 +1150,6 @@ defmodule ReqLLM.Provider.Defaults do
 
             if finish_reason do
               normalized_reason = parse_openai_finish_reason(finish_reason)
-
-              # The finish_reason chunk is deliberately NOT marked terminal:
-              # it is not the end of an OpenAI-compatible stream. With
-              # stream_options.include_usage (set automatically when
-              # streaming), a final usage chunk (choices: []) arrives *after*
-              # it, and marking finish_reason terminal finalized the stream
-              # early - dropping usage whenever it landed in a later network
-              # chunk. Termination is signaled by the standalone usage chunk
-              # below, the [DONE] event, or HTTP stream end.
               meta_chunk = ReqLLM.StreamChunk.meta(%{finish_reason: normalized_reason})
 
               content_chunks ++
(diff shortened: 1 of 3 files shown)


BinaryMuse:
Missed the bit about inline comments. Thanks!

mikehostetler:
@BinaryMuse No problem! Thanks for the PR!!!

PR #771, opened 2026-06-15
feat(bedrock): bidirectional streaming (Nova Sonic speech-to-speech)

neilberkman:
Adds support for AWS Bedrock `InvokeModelWithBidirectionalStream` — Amazon Nova Sonic speech-to-speech — over a true HTTP/2 full-duplex connection.

## What's here

- **`ReqLLM.Streaming.HTTP2DuplexSession`** — a generic Mint HTTP/2 duplex transport, mirroring the shape of `WebSocketSession`. It streams request DATA frames over time (with proper HTTP/2 flow control) while pushing response frames to its owner process, so reads and writes never block each other. The existing WebSockex transport is RFC6455 websockets and can't do HTTP/2, which this API requires.

- **`ReqLLM.Bedrock.BidiStream`** — `connect` / `send_event` / `next_event` / `close`, a session-oriented API in the same style as the repo's `ReqLLM.OpenAI.Realtime`. Each outbound event is framed as an AWS event-stream message (payload `{"bytes": base64(json)}`) wrapped in a SigV4 chunk-signed message (`AWS4-HMAC-SHA256-PAYLOAD`) chained from the initial request's seed signature. Inbound is decoded by the existing `AmazonBedrock.AWSEventStream` parser. `next_event` parks asynchronously, so one process can stream microphone audio while another consumes responses.

- **`ReqLLM.Bedrock.NovaSonic`** — event builders + a small orchestration API (`start`, `audio_turn`, `finish`, `next`) following the Nova Sonic v1 bidirectional event protocol.

## API shape

Bidirectional speech-to-speech is a live send-audio / receive-audio session, not a token stream, so it's exposed as a session API rather than through `stream_text/3` — exactly like the existing `OpenAI.Realtime` support. No changes to the high-level pipeline.

## Dependency

Requires `ex_aws_auth ~> 1.4` (1.4.1), which adds the event-stream request helpers this uses: `AWSAuth.EventStream.encode_string_header/2` and the `:streaming_events` seed option.

## Validation

Validated end-to-end against live Bedrock (`us-east-1`, IAM SigV4): real speech in → ASR transcription → assistant text + 24kHz audio out (confirmed intelligible). Unit tests cover the Nova Sonic event builders; the wire protocol (framing, chunk signing, flow control) is exercised by the live path. All CI green.


mikehostetler (changes requested):
Thanks for the substantial Bedrock/Nova Sonic work here. I found a few blocking issues to address before this is ready to merge:

- [P1] `lib/req_llm/bedrock/bidi_stream.ex:90` computes an override `region`, but `sign_seed/5` ignores it when calling `AWSAuth.sign_authorization_header/5`. If env credentials have `AWS_REGION=us-west-2` and the caller passes `region: "us-east-1"`, the host/event signatures use east but the seed Authorization scope uses west, so Bedrock auth fails. Pass `region: region` into the signer or normalize the credentials before signing.

- [P1] `lib/req_llm/bedrock/nova_sonic.ex:75` opens and closes a fresh audio content block for every `audio_turn/3`. AWS's Nova Sonic input flow describes continuous audio frames using the same content container until session cleanup, and the AWS example keeps one `audio_content_name` open until `end_session`. This implementation will not model continuous mic input, barge-in, or multi-turn audio correctly. The unused `audio_content_name` field looks like the right place to track that lifecycle.

- [P2] `test/req_llm/bedrock/nova_sonic_test.exs:1` only covers pure event builders. The new signing/session/transport code has no regression coverage, which is risky for 700+ lines of HTTP/2 and SigV4 behavior. Add focused tests for region override signing, event-stream frame construction, inbound dispatch/error handling, and the Nova Sonic audio lifecycle.

Validation checked: focused Nova Sonic tests pass, GitHub CI is green, and the PR is merge-clean. I’m marking this `needs_work` because the two behavior issues above are blocking.

neilberkman:
Thanks for the careful review — all three addressed, force-pushed (single commit), and re-validated end-to-end against live Bedrock with the new API.

**[P1] Region override (bidi_stream.ex)** — `sign_seed/5` now passes `region:` into `AWSAuth.sign_authorization_header/5`, so the seed Authorization scope matches the host and the per-event signatures even when the call overrides the credential region. Covered by a new test asserting the seed scopes to the override region and not the credential region.

**[P1] Continuous audio lifecycle (nova_sonic.ex)** — replaced the open/close-per-call model with `start_audio/2` → `send_audio/3` → `end_audio/1`, which keep a single audio content container open (tracked via `audio_content_name` on the session) so continuous mic input, barge-in, and multi-turn work within one block. `audio_turn/3` is now a thin one-shot convenience over those, and `finish/1` closes any open block before `promptEnd`/`sessionEnd`. Covered by a lifecycle test (one content name across many `send_audio` calls; error when none is open).

**[P2] Tests** — added focused unit tests beyond the event builders:
- region-override seed signing
- event-stream frame construction (`{"bytes": base64}` payload + CRC-valid prelude/message framing)
- inbound decode/dispatch (complete message, partial buffer, multiple concatenated messages)
- the audio content lifecycle

Also rebased on `main` to resolve the `mix.exs` conflict with the new `goth` dependency.

mikehostetler (approved):
Re-reviewed after the follow-up commit. The prior blockers are addressed, focused tests pass locally, CI is green, and the branch is merge-clean.

PR #780, opened 2026-06-18
fix(openai): capture trailing streamed usage chunk after finish_reason

### LINKED ISSUE #781: [Bug]: Streamed token usage lost on Azure/LiteLLM (usage chunk arrives after finish_reason)
### Provider (if applicable)
openai (Azure OpenAI, and OpenAI-compatible gateways such as LiteLLM)

### Model (if applicable)
Any chat model served via Azure OpenAI / LiteLLM (e.g. `gpt-4o`) with `stream_options: {include_usage: true}`

### Bug Description
For streaming chat completions, `ReqLLM.Response.usage/1` comes back with `input_tokens: 0` / `output_tokens: 0` (and therefore no cost) on Azure OpenAI and OpenAI-compatible gateways like LiteLLM — even though usage *is* being sent.

These providers stream token `usage` in a **separate SSE chunk that arrives _after_ the `finish_reason` chunk**, just before `[DONE]`:

```
data: {"choices":[{"finish_reason":"stop","index":0,"delta":{}}]}
data: {"choices":[{"index":0,"delta":{}}],"usage":{"prompt_tokens":12,"completion_tokens":7,...}}
data: [DONE]
```

`default_decode_stream_event/2` flags the `finish_reason` chunk `terminal?: true`, so the stream finalizes there and a consumer reading `Response.usage` right after "completion" races the still-in-flight usage chunk. Non-streaming responses are unaffected (usage is in the single body). In practice it's deterministic — the usage chunk is a separate network frame, so the read almost always wins the race.

### Reproduction Code
```elixir
{:ok, stream} =
  ReqLLM.stream_text(
    "openai:gpt-4o",
    "say hi in three words",
    provider_options: [stream_options: %{include_usage: true}]
  )

{:ok, response} = ReqLLM.StreamResponse.to_response(stream)
ReqLLM.Response.usage(response)
# Against Azure OpenAI / LiteLLM:
#   %{input_tokens: 0, output_tokens: 0, ...}   # expected non-zero
```

### Expected Behavior
`Response.usage` reflects the gateway-reported token counts (e.g. `input_tokens: 12, output_tokens: 7`), as it does for non-streaming responses.

### Actual Behavior
`input_tokens: 0, output_tokens: 0` — the trailing usage chunk is dropped because the stream finalizes on `finish_reason` before that chunk is accumulated.

### Environment
- ReqLLM: `main`
- Elixir: 1.19
- OTP: 27

valentin-ib:
## Description

With `stream_options: {include_usage: true}`, Azure OpenAI and OpenAI-compatible gateways (e.g. LiteLLM) send token `usage` in a **separate SSE chunk that arrives _after_ the `finish_reason` chunk**, just before `[DONE]`:

```
data: {"choices":[{"finish_reason":"stop","index":0,"delta":{}}]}
data: {"choices":[{"index":0,"delta":{}}],"usage":{"prompt_tokens":12,"completion_tokens":7,...}}
data: [DONE]
```

`default_decode_stream_event/2` flags the `finish_reason` chunk `terminal?: true`, so `StreamServer` finalizes the stream there. A consumer that reads `ReqLLM.Response.usage/1` as soon as the stream "completes" then **races the still-in-flight usage chunk** and gets `input_tokens: 0 / output_tokens: 0` (and therefore no cost). Non-streaming responses are unaffected — usage is in the single body. In practice this is deterministic: the usage chunk is a separate network frame, so the read almost always wins the race.

### Fix

`ReqLLM.Providers.OpenAI.ChatAPI.decode_stream_event/2` strips the `terminal?` flag off **normal-completion** `finish_reason` chunks, so the stream finalizes on `[DONE]` (or connection close) instead — by which point the trailing usage chunk has been accumulated into metadata.

Guards keep existing behavior intact:

- **Inline error chunks stay terminal.** `finish_reason: :error` (and any meta carrying an `:error` key — how OpenAI-compatible gateways report mid-stream failures via `data: {"error": ...}`) keep `terminal?`, so failures still surface immediately instead of waiting for a `[DONE]` that won't come.
- `[DONE]` and empty-`choices` usage chunks have no `:finish_reason` key, so they're untouched and keep their own terminal flag.

Scoped to the OpenAI **ChatAPI** driver; non-streaming and other providers are unaffected.

### Alternative considered

The deeper root cause is that `StreamServer` snapshots metadata at the terminal chunk, before the trailing usage chunk merges. A fix there (defer finalization/metadata until the stream truly ends) would be more general but a larger change to the finalization lifecycle. This PR is the minimal, decoder-local fix — happy to take the broader approach instead if you'd prefer.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None. The public API and the `Response` struct are unchanged; streamed `Response.usage` is simply populated where it was previously zero.

## Testing

- [x] Tests pass (`mix test`)
- [x] Quality checks pass (`mix format` / compile clean; CI runs full `mix quality`)

New unit tests in `test/providers/openai_chat_streaming_usage_test.exs` (pure decode tests — no live API / fixtures needed):

- `finish_reason` chunk is no longer terminal
- the trailing usage chunk yields usage
- a combined `finish_reason` + `usage` chunk still yields usage
- an empty-`choices` usage chunk keeps its terminal flag
- `[DONE]` stays terminal
- an inline error chunk stays terminal

## Checklist

- [x] I have added tests that prove my fix works
- [x] I have **NOT** edited `CHANGELOG.md` (auto-generated by git_ops)

## Related Issues

Closes #781

### DIFF
--- a/lib/req_llm/providers/openai/chat_api.ex
+++ b/lib/req_llm/providers/openai/chat_api.ex
@@ -70,33 +70,7 @@ defmodule ReqLLM.Providers.OpenAI.ChatAPI do
 
   @impl true
   def decode_stream_event(event, model) do
-    event
-    |> ReqLLM.Provider.Defaults.default_decode_stream_event(model)
-    |> drop_finish_reason_terminal()
-  end
-
-  # Azure and OpenAI-compatible gateways (e.g. LiteLLM) send the usage chunk
-  # after the finish_reason chunk, just before [DONE]. The default decoder marks
-  # finish_reason terminal, so the stream ends before that usage chunk lands and
-  # Response.usage comes back zero. Drop terminal? on normal finish_reason chunks
-  # so the stream finalizes on [DONE] (or connection close) once usage is in.
-  # Error chunks (finish_reason: :error, or an :error key) stay terminal so
-  # failures still fail fast; [DONE] and empty-choices usage chunks have no
-  # finish_reason and are left untouched.
-  defp drop_finish_reason_terminal(chunks) do
-    Enum.map(chunks, fn
-      %ReqLLM.StreamChunk{
-        type: :meta,
-        metadata: %{terminal?: true, finish_reason: reason} = meta
-      } = chunk
-      when reason != :error ->
-        if Map.has_key?(meta, :error),
-          do: chunk,
-          else: %{chunk | metadata: Map.delete(meta, :terminal?)}
-
-      chunk ->
-        chunk
-    end)
+    ReqLLM.Provider.Defaults.default_decode_stream_event(event, model)
   end
 
   # ========================================================================
--- a/lib/req_llm/providers/openai/chat_api.ex
+++ b/lib/req_llm/providers/openai/chat_api.ex
@@ -70,7 +70,33 @@ defmodule ReqLLM.Providers.OpenAI.ChatAPI do
 
   @impl true
   def decode_stream_event(event, model) do
-    ReqLLM.Provider.Defaults.default_decode_stream_event(event, model)
+    event
+    |> ReqLLM.Provider.Defaults.default_decode_stream_event(model)
+    |> drop_finish_reason_terminal()
+  end
+
+  # Azure and OpenAI-compatible gateways (e.g. LiteLLM) send the usage chunk
+  # after the finish_reason chunk, just before [DONE]. The default decoder marks
+  # finish_reason terminal, so the stream ends before that usage chunk lands and
+  # Response.usage comes back zero. Drop terminal? on normal finish_reason chunks
+  # so the stream finalizes on [DONE] (or connection close) once usage is in.
+  # Error chunks (finish_reason: :error, or an :error key) stay terminal so
+  # failures still fail fast; [DONE] and empty-choices usage chunks have no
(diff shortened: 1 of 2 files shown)


valentin-ib:
For context, I identified this whilst setting up a token and cost usage dashboard, and noticed that the usage wasn't coming through (it was always 0). So wanted to raise it and check if anyone else thinks this is a reasonable amendment.

mikehostetler:
Thanks for the detailed repro. I dug into this against the latest PR head: after rebasing, the decoder-level claim in the original description no longer matches current `main` because `default_decode_stream_event/2` already emits the `finish_reason` meta chunk without `terminal?`.

I removed the redundant `ChatAPI` override and added an end-to-end regression that replays the ordering from #781: content, then `finish_reason: "stop"`, then a separate usage chunk with non-empty `choices`, then `[DONE]`. That test verifies `StreamResponse.to_response/1` returns the trailing usage as non-zero `Response.usage`.

Could you please test the latest PR branch against your original Azure OpenAI / LiteLLM setup or token/cost dashboard repro? The simulated SSE path is covered and CI is green, but I want confirmation that the real provider path that produced zero usage is fixed before we treat #781 as fully closed.

PR #801, opened 2026-07-03
Fix: streaming tool-call arguments dropped when the leading brace ships with the tool name

anagrius:
## Problem

Some OpenAI-compatible servers (**llama.cpp**, **vLLM**, …) begin streaming a tool call's `arguments` in the **same** chunk as the function `name`. This is valid per the OpenAI streaming spec — the first delta looks like:

```json
{"index":0,"id":"…","type":"function","function":{"name":"list_metrics","arguments":"{"}}
```

…and the continuation deltas carry the rest:

```json
{"index":0,"function":{"arguments":"\"database\":\""}}
{"index":0,"function":{"arguments":"dev"}}
…
{"index":0,"function":{"arguments":"\"}"}}
```

`decode_openai_tool_call_delta/1` matches the first chunk on the "complete tool call" clause (name **and** arguments present), calls `build_openai_tool_call_chunk("list_metrics", "{", …)`, `Jason.decode("{")` fails, and the `"{"` is preserved only as `:raw_arguments` metadata on the tool-call chunk — it is **never** added to the per-index argument-fragment buffer.

The continuation fragments then accumulate to `"database":"dev","table":"dev"}` — valid JSON **except the leading brace is missing** — so `message_tool_call_args/2` fails to decode and falls back to the tool-call chunk's own (empty) arguments. **The tool ends up being called with `%{}`.**

OpenAI/Anthropic/etc. are unaffected because their first tool-call chunk sends `arguments: ""` (empty), so every argument byte arrives via the continuation path.

## Fix

Seed the per-index fragment buffer with the tool-call chunk's `:raw_arguments` (the dropped leading fragment) in `ChunkAccumulator.push/2` for `:tool_call` chunks. The continuation fragments then append to it and the joined JSON parses correctly. Providers that send an empty `arguments` in the name chunk have an empty `raw_arguments`, so nothing is seeded and their behavior is unchanged.

## Test

Added a regression test in `chunk_accumulator_test.exs` covering the llama.cpp/vLLM shape (first chunk carries `"{"`, continuation fragments carry the rest). Verified it fails without the fix and passes with it; the full `ChunkAccumulator` suite is green.


anagrius:
Without this the lib is broken for many custom providers.

PR #825, opened 2026-07-15
fix(streaming): terminate metadata handles

### LINKED ISSUE #824: [Bug]: MetadataHandle GenServers never terminate after stream processing
### Provider (if applicable)

Google Gemini, although the leak is provider-independent.

### Model (if applicable)

Observed with Gemini streaming requests.

## Summary

Every call to `ReqLLM.stream_text/3` starts a `ReqLLM.StreamResponse.MetadataHandle` GenServer. After the stream has been fully consumed and its metadata has been materialized by `ReqLLM.StreamResponse.process_stream/2`, the handle remains alive indefinitely in the ready state.

There is no public cleanup operation, owner monitoring, idle timeout, or self-termination path. In a long-running BEAM node this produces one permanent process per streaming request.

This appears analogous to #492, which reported the same lifecycle problem for `StreamServer` and was fixed by #494. That fix did not cover `MetadataHandle`.

## Production evidence

Observed in a long-running Phoenix application using ReqLLM 1.5.1:

- 34,848 of 36,060 BEAM processes were `ReqLLM.StreamResponse.MetadataHandle` processes.
- Sampled handles were all in `metadata: {:ready, metadata}` state.
- They had `waiters: []`, `fetch_fun: nil`, empty mailboxes, and no links.
- 18,048 LLM turns had occurred since the last deployment, approximately 1.93 surviving handles per turn because some turns perform multiple tool-loop requests.
- Pod memory grew steadily from roughly 3.4 GiB to 5.8 GiB over thirteen days.

The implementation is unchanged between v1.5.1, v1.17.1, and current `main` (the file has the same blob SHA in all three).

After metadata collection completes, the server returns:

```elixir
{:noreply, %{state | metadata: {:ready, metadata}, waiters: [], fetch_fun: nil}}
```

Subsequent `:await` calls continue returning the cached value, but nothing ever terminates the process.

### Reproduction Code

Using a real streaming request:

```elixir
handles =
  for _ <- 1..100 do
    {:ok, stream_response} =
      ReqLLM.stream_text(model, context, request_options)

    {:ok, _response} =
      ReqLLM.StreamResponse.process_stream(stream_response)

    stream_response.metadata_handle
  end

Enum.count(handles, &Process.alive?/1)
# => 100
```

The lower-level lifecycle can also be demonstrated without a provider:

```elixir
alias ReqLLM.StreamResponse.MetadataHandle

handles =
  for _ <- 1..100 do
    {:ok, handle} =
      MetadataHandle.start_link(fn ->
        %{usage: %{input_tokens: 1}, finish_reason: :stop}
      end)

    _metadata = MetadataHandle.await(handle)
    Process.unlink(handle)
    handle
  end

Enum.count(handles, &Process.alive?/1)
# => 100
```

### Expected Behavior

Completed streaming operations should have a deterministic way to release their metadata process.

In particular:

- Terminal materialization APIs such as `process_stream/2` and `to_response/1` should not leave an unreachable metadata process behind after copying metadata into a plain `Response`.
- Callers that access `usage/1` and `finish_reason/1` directly should have an explicit lifecycle API such as `StreamResponse.close/1` or `MetadataHandle.stop/1`.
- Repeated metadata reads should remain supported until the handle is closed.
- Normal completion, cancellation, errors, and consumer termination should all have tested cleanup behavior.

### Actual Behavior

Each completed stream leaves a live `MetadataHandle` GenServer holding the cached metadata indefinitely. Dropping the `StreamResponse` struct does not release the process, and normal exit of the linked creator does not terminate it.

### Suggested Direction

One possible lifecycle contract:

1. Add an idempotent public `StreamResponse.close/1` operation that terminates the metadata handle safely.
2. Have terminal consuming APIs such as `process_stream/2` and `to_response/1` close the handle in an `after` block once the metadata has been copied into the returned `Response`.
3. Preserve the current reusable metadata behavior for callers using `usage/1` and `finish_reason/1` until they explicitly close the response.
4. Add a regression test that processes many streams and asserts their metadata handles terminate, including error and cancellation paths.

An ownership-monitoring design could also work, provided transferring a `StreamResponse` between processes remains safe.

### Environment

- Confirmed in ReqLLM 1.5.1
- Source inspection confirms the same `MetadataHandle` implementation in ReqLLM 1.17.1 and current `main`
- Long-running Elixir/Phoenix application on Kubernetes

mikehostetler:
## Summary

- add idempotent lifecycle APIs for closing `StreamResponse` and stopping `MetadataHandle`
- close metadata handles after `process_stream/2`, `to_response/1`, explicit cancellation, and partial stream consumption
- keep metadata reads reusable until callers explicitly close non-materialized responses
- make metadata collection independently stoppable so a blocked collector cannot prevent cleanup

## Root cause

`MetadataHandle` cached completed metadata in a GenServer indefinitely. Terminal materialization copied that metadata into a `Response` but never stopped the GenServer, while cancellation and stream resource cleanup only terminated the `StreamServer`.

## Impact

Completed, errored, cancelled, and partially consumed streams no longer retain permanent metadata processes. Existing callers can continue reading usage and finish reason repeatedly until they call `StreamResponse.close/1`.

## Validation

- `mix test` (3,566 passed, 11 skipped)
- `mix test test/req_llm/stream_response_test.exs` (96 passed)
- `mix test test/req_llm/integration/streaming_orchestration_test.exs --include integration` (11 passed)
- `mix quality`

Closes #824

### DIFF
--- a/lib/req_llm.ex
+++ b/lib/req_llm.ex
@@ -942,8 +942,9 @@ defmodule ReqLLM do
   ## Performance Notes
 
   The stream is lazy and supports backpressure. Metadata collection happens
-  concurrently and won't block token delivery. Use cancellation for early
-  termination to free resources. High-concurrency streaming workloads can tune
+  concurrently and won't block token delivery. Use `ReqLLM.StreamResponse.close/1`
+  for early termination or after direct metadata access to free resources.
+  High-concurrency streaming workloads can tune
   Finch pool protocols with `:stream_pool_protocols`, capacity with
   `:stream_pool_size` and `:stream_pool_count`, shard selection with
   `:stream_pool_strategy`, and checkout behavior with `:stream_pool_timeout`
--- a/lib/req_llm/stream_response.ex
+++ b/lib/req_llm/stream_response.ex
@@ -32,6 +32,7 @@ defmodule ReqLLM.StreamResponse do
 
   text = ReqLLM.StreamResponse.text(stream_response)
   usage = ReqLLM.StreamResponse.usage(stream_response)
+  :ok = ReqLLM.StreamResponse.close(stream_response)
   ```
 
   ### Backward compatibility
@@ -52,8 +53,8 @@ defmodule ReqLLM.StreamResponse do
   |> Stream.each(&IO.write/1)
   |> Stream.run()
 
-  # Cancel remaining work
-  stream_response.cancel.()
+  # Cancel remaining work and close the metadata handle
+  ReqLLM.StreamResponse.close(stream_response)
   ```
 
   ## Design Philosophy
@@ -385,6 +386,7 @@ defmodule ReqLLM.StreamResponse do
   - The stream is consumed exactly once (no double-consumption bugs)
   - All callbacks are optional - omitted callbacks are simply not invoked
   - The returned Response struct contains all accumulated data plus metadata
+  - The metadata handle is closed before this function returns
   """
   @spec process_stream(t(), keyword()) :: {:ok, Response.t()} | {:error, term()}
   def process_stream(%__MODULE__{} = stream_response, opts \\ []) do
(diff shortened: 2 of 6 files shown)


javiercr:
Thank you!

PR #936, opened 2026-07-30
fix(google): preserve raw Gemini finishReason and finishMessage in stream meta

### LINKED ISSUE #935: [Bug]: google: unrecognized finishReason (e.g. MALFORMED_FUNCTION_CALL) collapsed to "error", finishMessage discarded
### What happened?

`normalize_google_finish_reason/1` in the Google provider maps only `STOP`, `MAX_TOKENS`, `SAFETY`, and `RECITATION`; every other Gemini `finishReason` — `MALFORMED_FUNCTION_CALL`, `UNEXPECTED_TOOL_CALL`, `BLOCKLIST`, `PROHIBITED_CONTENT`, `OTHER`, … — is collapsed to `"error"`. The stream decoder (`decode_google_event/2`) also drops Google's explanatory `finishMessage` entirely. Neither the original reason string nor the message survives into the meta chunk or anywhere else, so downstream consumers cannot tell *why* a generation aborted.

### Real-world impact

We chased a production-grade bug where `gemini-3.5-flash` deterministically aborted an agent continuation a few dozen thinking-tokens in. All our telemetry could say was "stream finished with error", and our app retried the identical request five times (it could never succeed — the failure is deterministic for the context). Only by capturing the raw SSE off-pipeline did we learn the actual cause:

```
"finishReason": "MALFORMED_FUNCTION_CALL",
"finishMessage": "Malformed function call: Failed to parse function call: Function call is empty - no input to parse."
```

With the reason visible, the right handling is obvious (e.g. don't blind-retry; raise the thinking level). Without it, every distinct abort mode is the same opaque error.

### Relation to prior work

This is the Google-provider sibling of #749 (`anthropic: map refusal/pause_turn/... stop_reasons instead of collapsing to :error`) — same lost-signal problem. Unlike the Anthropic reasons, most of Gemini's abort reasons have no good canonical `finish_reason` equivalent, so rather than remapping, the fix preserves the raw values alongside the normalized one.

### Fix

PR incoming: keep `finish_reason` normalization unchanged, and additionally carry `finish_reason_raw` (always) and `finish_message` (when present) in the terminal meta chunk.

### Library version

1.17.1 (also present on current main)

gi11es:
## Description

`normalize_google_finish_reason/1` collapses every unrecognized Gemini `finishReason` (`MALFORMED_FUNCTION_CALL`, `UNEXPECTED_TOOL_CALL`, `BLOCKLIST`, `PROHIBITED_CONTENT`, `OTHER`, …) into `"error"`, and `decode_google_event/2` discards Google's explanatory `finishMessage`. Downstream consumers can only see an opaque error, which makes deterministic provider-side aborts undiagnosable — we hit this in production with `gemini-3.5-flash` aborting on `MALFORMED_FUNCTION_CALL` ("Function call is empty - no input to parse") and could only observe "stream finished with error" while retrying a request that could never succeed.

This is the Google sibling of #749 (Anthropic stop_reasons collapsed to `:error`). Unlike those, most Gemini abort reasons have no good canonical `finish_reason` mapping, so instead of remapping, this change leaves normalization untouched and additionally carries in the terminal meta chunk:

- `finish_reason_raw` — the original `finishReason` string, always;
- `finish_message` — Google's `finishMessage`, when present.

Happy to nest these under `provider_meta` (like grounding metadata) instead of top-level meta keys if that's the preferred shape.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None — existing `finish_reason` values are unchanged; the new keys are additive.

## Testing

- [x] Tests pass (`mix test`, 4040 passed)
- [x] Quality checks pass (`mix quality`, dialyzer clean)

Added three decode tests: abnormal finish with `finishMessage` (the MALFORMED_FUNCTION_CALL case, raw values preserved and normalization still `"error"`), abnormal finish without `finishMessage` (key absent), and STOP (raw carried, normalization unchanged). `mix mc google` unchanged (2/31 models covered locally; decode-only change).

## Checklist

## Related Issues

Closes #935

### DIFF
--- a/lib/req_llm/providers/google.ex
+++ b/lib/req_llm/providers/google.ex
@@ -1902,6 +1902,25 @@ defmodule ReqLLM.Providers.Google do
 
   defp attach_reasoning_details(response, _details), do: response
 
+  # Preserve the provider's original finishReason (and finishMessage, when
+  # present) alongside the normalized value. Normalization collapses every
+  # unrecognized reason (MALFORMED_FUNCTION_CALL, UNEXPECTED_TOOL_CALL,
+  # OTHER, ...) into "error", which makes provider failures undiagnosable
+  # downstream.
+  defp put_raw_finish_meta(meta, finish_reason, data) do
+    finish_message =
+      case data do
+        %{"candidates" => [%{"finishMessage" => message} | _]} when is_binary(message) -> message
+        _ -> nil
+      end
+
+    meta
+    |> Map.put(:finish_reason_raw, finish_reason)
+    |> then(fn m ->
+      if finish_message, do: Map.put(m, :finish_message, finish_message), else: m
+    end)
+  end
+
   defp normalize_google_finish_reason("STOP"), do: "stop"
   defp normalize_google_finish_reason("MAX_TOKENS"), do: "length"
   defp normalize_google_finish_reason("SAFETY"), do: "content_filter"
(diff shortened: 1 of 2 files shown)


mikehostetler (approved):
Thank you for this change. I found no blocking issues. The change preserves Google's raw finish reason and message in terminal stream metadata without changing the normalized finish reason. The tests cover normal and abnormal stops, including a missing finish message. The Google provider test file passes locally (114 tests), the PR is clean against main, and all checks are green.

PR #939 edits the same test area, so the second PR to merge will need a small conflict resolution.

PR #939, opened 2026-07-31
fix(google): map Gemini policy stops to content_filter

### LINKED ISSUE #938: [Bug]: google: policy stops (BLOCKLIST, PROHIBITED_CONTENT, SPII, LANGUAGE, IMAGE_*) normalize to "error" instead of "content_filter"
### What happened?

`normalize_google_finish_reason/1` maps `SAFETY` and `RECITATION` to `"content_filter"`, but Gemini has several sibling finish reasons for the same outcome — the response candidate was flagged — and none of them are mapped:

```
BLOCKLIST, PROHIBITED_CONTENT, SPII, LANGUAGE,
IMAGE_SAFETY, IMAGE_PROHIBITED_CONTENT, IMAGE_RECITATION
```

They fall through the catch-all and surface as `"error"`.

The `FinishReason` enum in `google/ai/generativelanguage/v1beta/generative_service.proto` describes this whole group with the same phrasing — "the response candidate content was flagged for ..." — so they are one outcome wearing different labels, and `SAFETY` is simply the one that got mapped.

### Why it matters

`"error"` reads as a transient provider fault, so callers retry. These stops are deterministic: the same request will be flagged again, and every retry burns quota and latency to reach the identical outcome. `"content_filter"` already exists as the canonical value for exactly this, and the two mapped siblings already use it.

### Expected behavior

All the content-flag finish reasons normalize to `"content_filter"`, consistent with `SAFETY`/`RECITATION`.

Note that some Gemini stops are *not* content flags and should not be swept in: `UNEXPECTED_TOOL_CALL` (a tool call with no tools enabled — a request-construction fault), `TOO_MANY_TOOL_CALLS` (system-aborted runaway), `NO_IMAGE`, `IMAGE_OTHER`, `OTHER`. Retrying most of those is equally futile, but "the model refused" would be the wrong story for them, so they should keep the existing behaviour until there is a non-retryable, non-refusal classification to give them.

### Prior art

This is the Google-provider sibling of #749 (`anthropic: map refusal/pause_turn/model_context_window_exceeded stop_reasons instead of collapsing to :error`) — same class of lost signal, same fix shape.

### Library version

1.17.1 (also present on current main)

### Fix

PR incoming.

gi11es:
## Description

`normalize_google_finish_reason/1` maps `SAFETY` and `RECITATION` to `"content_filter"`, but the sibling finish reasons for the same outcome — `BLOCKLIST`, `PROHIBITED_CONTENT`, `SPII`, `LANGUAGE`, `IMAGE_SAFETY`, `IMAGE_PROHIBITED_CONTENT`, `IMAGE_RECITATION` — fall through the catch-all and become `"error"`. The `FinishReason` proto describes the whole group in the same words ("the response candidate content was flagged for ..."), so they are one outcome wearing different labels.

Reporting them as `"error"` makes a deterministic, non-retryable stop look like a transient provider fault, so callers retry a request that can never succeed.

This is the Google sibling of #749 (Anthropic stop_reasons collapsed to `:error`).

**Deliberately not mapped**, and pinned by a test so a later tidy-up pass doesn't sweep the whole enum in: `UNEXPECTED_TOOL_CALL` (tool call with no tools enabled — a request-construction fault), `TOO_MANY_TOOL_CALLS` (system-aborted runaway), `NO_IMAGE`, `IMAGE_OTHER`, `OTHER`. Retrying most of these is equally futile, but they need a non-retryable non-refusal classification that doesn't exist yet, and "the model refused" would be the wrong story for them.

A note on sourcing the names, since no single reference is complete: the values were taken from the `FinishReason` enum in `googleapis/googleapis` (`google/ai/generativelanguage/v1beta/generative_service.proto`) and cross-checked against both generated clients — the public REST reference omits the `IMAGE_*` values, and google-genai's Python types omit `TOO_MANY_TOOL_CALLS`.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None in the API surface. Behaviour does change for the seven listed finish reasons: `finish_reason` becomes `"content_filter"` where it was previously `"error"` — which is the point of the fix, and matches how `SAFETY`/`RECITATION` already behave. Unmapped values are untouched.

## Testing

- [x] Tests pass (`mix test`, 4039 passed, 11 skipped)
- [x] Quality checks pass (`mix quality`, dialyzer clean, credo --strict clean)

Two new decode tests: one asserting all seven content-flag reasons normalize to `"content_filter"`, one pinning the five non-content-flag stops as `"error"`. Both fail without the lib change.

`mix mc "google:*"` is byte-identical before and after the change (verified by running it on unpatched `main` and on this branch):

```
Summary: 2/31 active models tested (6.5% coverage)
Overall Coverage: 2/31 active models passing (6.5%) in 23s
```

The low coverage is local fixture availability, not a regression — most Google fixtures aren't in the repo. Happy to re-run any specific model if you can point me at the fixtures.

## Checklist

## Related Issues

Closes #938

Independent of #936 (raw finishReason preservation) — they touch the same function but neither depends on the other, and they can merge in either order.

### DIFF
--- a/lib/req_llm/providers/google.ex
+++ b/lib/req_llm/providers/google.ex
@@ -1906,6 +1906,35 @@ defmodule ReqLLM.Providers.Google do
   defp normalize_google_finish_reason("MAX_TOKENS"), do: "length"
   defp normalize_google_finish_reason("SAFETY"), do: "content_filter"
   defp normalize_google_finish_reason("RECITATION"), do: "content_filter"
+
+  # Gemini flags generated content under several names beyond SAFETY. The proto
+  # describes this whole group in the same words — "the response candidate
+  # content was flagged for ..." — so they are one outcome wearing different
+  # labels and normalize the same way. Collapsing them into "error" instead
+  # makes a deterministic, non-retryable stop look like a transient provider
+  # fault, which callers then retry pointlessly.
+  #
+  # Source of truth is the FinishReason enum in googleapis/googleapis
+  # (google/ai/generativelanguage/v1beta/generative_service.proto). Cross-check
+  # both generated clients before adding names: the public REST reference omits
+  # the IMAGE_* values, and google-genai's Python types omit
+  # TOO_MANY_TOOL_CALLS, so neither is complete on its own.
+  #
+  # Deliberately NOT mapped here, because they are not content flags and want
+  # different handling than "the model refused": UNEXPECTED_TOOL_CALL (a tool
+  # call with no tools enabled in the request — a request-construction fault),
+  # TOO_MANY_TOOL_CALLS (the system aborted a tool-call runaway), NO_IMAGE /
+  # IMAGE_OTHER, and OTHER. Retrying most of these is equally futile, but they
+  # need a non-retryable non-refusal classification that does not exist yet, so
+  # they keep the existing "error" behaviour rather than borrow the wrong one.
+  defp normalize_google_finish_reason("BLOCKLIST"), do: "content_filter"
+  defp normalize_google_finish_reason("PROHIBITED_CONTENT"), do: "content_filter"
+  defp normalize_google_finish_reason("SPII"), do: "content_filter"
+  defp normalize_google_finish_reason("LANGUAGE"), do: "content_filter"
+  defp normalize_google_finish_reason("IMAGE_SAFETY"), do: "content_filter"
+  defp normalize_google_finish_reason("IMAGE_PROHIBITED_CONTENT"), do: "content_filter"
+  defp normalize_google_finish_reason("IMAGE_RECITATION"), do: "content_filter"
+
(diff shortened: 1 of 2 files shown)


mikehostetler (approved):
Thank you for this change. I found no blocking issues. The new mappings match Google's finish-reason descriptions, and the tests cover both mapped and deliberately unmapped reasons. The Google provider test file passes locally (113 tests), the PR is clean against main, and all checks are green.

Non-blocking note: the current REST reference lists a few newer reasons that are absent from the proto. The default clause handles them safely. PR #936 edits the same test area, so the second PR to merge will need a small conflict resolution.

mikehostetler (approved):
Re-approved after the conflict resolution. Both test groups are present, 116 Google provider tests pass locally, and the full CI run is green.

PR #942, opened 2026-07-31
feat(openai): support explicit cache breakpoints

### LINKED ISSUE #937: [Feature]: Explicit cache breakpoints
### Feature Type

Provider feature extension

### Provider (if applicable)

openai

### Description

Hello, this is part question part feature request. If it's not currently supported I don't mind giving implementation a try (I didn't dive into low level message assembly before it's sent to AI providers so it could take some time 😆 ).

I'm doing agressive fanout for one of my AI jobs and it seems like there's no way to explicitly set caching breakpoints for openAI providers right? Ideally I think that there'd be 'provider agnostic' way of doing that (I'm at a phase where I still change models/providers frequently and ideally that would be done without thinking about changing caching way), but that's just 'nice to have'. My main request is for OpenAI explicit caching to be added.

### Example Usage

For now something like that would be sufficient.

```elixir
ReqLLM.Context.user([
    ContentPart.text(
      "SOURCE MATERIAL:\n...",
      %{
        prompt_cache_breakpoint: %{mode: "explicit"}
      }
    ),
    ContentPart.text(
      "More text that should remain dynamic"
    )
  ])
``` 

The ideal usage would probably be some `ReqLLM.Context.ContentPart.cache_breakpoint()` that would be totally provider agnostic and maybe be typechecked given new type system.

### Can you contribute this?

Yes, I can submit a PR

mikehostetler:
## Summary

- pass `prompt_cache_breakpoint` metadata through supported text, image, and file content blocks
- add `prompt_cache_key` and `prompt_cache_options` provider options
- support both Chat Completions and Responses request formats
- normalize OpenAI `cache_write_tokens` as `cache_creation_tokens`
- document the GPT-5.6+ usage pattern

## Impact

Applications with stable prompt prefixes can select explicit OpenAI cache boundaries and inspect both cache reads and cache writes. The new request fields are intended for GPT-5.6 and later models; older models reject explicit cache controls.

The implementation stays OpenAI-specific because provider cache policies use different request semantics.

## Validation

- focused OpenAI/default encoder suites: 211 passed
- end-to-end provider option and request encoding check
- `mix quality`
- `mix test` (4,044 of 4,046 passed; two unrelated transient failures passed with `mix test --failed`)

## Reference

- [OpenAI prompt cache breakpoints](https://developers.openai.com/api/docs/guides/prompt-caching#prompt-cache-breakpoints)

Closes #937

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -1028,19 +1028,22 @@ defmodule ReqLLM.Provider.Defaults do
 
   defp encode_openai_content_part(%ReqLLM.Message.ContentPart{
          type: :file,
-         file_id: file_id
+         file_id: file_id,
+         metadata: metadata
        })
        when is_binary(file_id) and file_id != "" do
     %{
       type: "file",
       file: %{file_id: file_id}
     }
+    |> merge_content_metadata(metadata)
   end
 
   defp encode_openai_content_part(%ReqLLM.Message.ContentPart{
          type: :file,
          data: data,
-         media_type: media_type
+         media_type: media_type,
+         metadata: metadata
        })
        when is_binary(data) do
     # Encode file as image_url data URI (OpenAI format supports various media types this way)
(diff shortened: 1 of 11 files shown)


nxy7:
Thanks for implementing this 🔥

PR #945, opened 2026-08-02
fix(google): stop reporting non-streaming aborts as a clean stop

### LINKED ISSUE #944: [Bug]: google: non-streaming abort with no content is reported as finish_reason "stop"
### What happened?

On the non-streaming endpoint, a Gemini candidate can arrive carrying only a `finishReason` and **no `content` key at all** — this is what Gemini sends when it aborts before producing anything. `UNEXPECTED_TOOL_CALL` reproduces roughly 9 times out of 10 against `gemini-2.5-pro` with no tools declared, and `MALFORMED_FUNCTION_CALL` turns up there too.

`convert_google_to_openai_format/1` has clauses for a candidate with `content.parts` and for one with `content` but no parts. A candidate with neither matches neither clause, so it falls through to the catch-all, which hardcodes:

```elixir
%{"message" => %{"role" => "assistant", "content" => ""}, "finish_reason" => "stop"}
```

### Why it matters

A provider abort becomes indistinguishable from an empty-but-successful answer. The caller sees `finish_reason: :stop` with empty content and has no way to tell that generation failed — and because the reason is discarded before normalization, it isn't recoverable downstream either.

The streaming path has preserved the raw reason since #936; this closes the same hole on the non-streaming path, which is the one tool-less requests (e.g. image turns) actually use.

### Expected behavior

The finish reason is normalized like every other (so an abort reports `"error"`, or `"content_filter"` for the policy stops added in #939), and the raw value is preserved so callers can distinguish `UNEXPECTED_TOOL_CALL` from any other failure.

### Library version

1.18.0

### Fix

PR incoming: adds the missing clause, plus a non-streaming counterpart to `put_raw_finish_meta/3` that carries `finish_reason_raw` (and `finish_message` when present) in `provider_meta`.

gi11es:
## Description

On the non-streaming endpoint a Gemini candidate can arrive carrying only a `finishReason`, with no `content` key at all — Gemini does this when it aborts before producing anything. `UNEXPECTED_TOOL_CALL` reproduces ~9 times out of 10 against `gemini-2.5-pro` with no tools declared, and `MALFORMED_FUNCTION_CALL` turns up there too.

That shape matched neither content-bearing clause in `convert_google_to_openai_format/1`, so it fell through to the catch-all and was reported as `finish_reason: "stop"`: a provider abort was indistinguishable from an empty but successful answer, and the raw reason was discarded before any caller could see it.

This adds the missing clause so the reason is normalized like every other, and carries `finish_reason_raw` (plus `finish_message` when present) in `provider_meta` — the non-streaming counterpart of `put_raw_finish_meta/3`. The streaming path has preserved the raw value since #936; this closes the same hole on the path tool-less requests actually use.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None in the API surface. Behaviour changes for the specific case that was previously mis-reported: a content-less candidate with a `finishReason` now yields the normalized reason (`"error"`, or `"content_filter"` for the policy stops from #939) instead of `"stop"`. Candidates that carry content are untouched.

## Testing

- [x] Tests pass (`mix test`, 4051 passed, 11 skipped)
- [x] Quality checks pass (`mix quality`, dialyzer clean, credo --strict clean)

New test covers the content-less abort candidate: normalized reason plus `finish_reason_raw`/`finish_message` in `provider_meta`. It fails without the lib change.

`mix mc "google:*"` is unchanged from `main` (2/33 active models locally — fixture availability, not a regression):

```
Summary: 2/33 active models tested (6.1% coverage)
Overall Coverage: 2/33 active models passing (6.1%) in 23s
```

## Checklist

## Related Issues

Closes #944

Completes the Google finish-reason work from #936 (streaming raw reason) and #939 (policy stop mapping) by covering the non-streaming path.

### DIFF
--- a/lib/req_llm/providers/google.ex
+++ b/lib/req_llm/providers/google.ex
@@ -1761,12 +1761,6 @@ defmodule ReqLLM.Providers.Google do
             "finish_reason" => normalize_google_finish_reason(finish_reason)
           }
 
-        # A candidate can carry a finishReason and no "content" key at all:
-        # Gemini does this when it aborts before producing anything, e.g.
-        # UNEXPECTED_TOOL_CALL or MALFORMED_FUNCTION_CALL on the non-streaming
-        # endpoint. Without this clause it fell through to the catch-all below
-        # and was reported as a clean "stop", so a provider abort was
-        # indistinguishable from an empty but successful answer.
         %{"finishReason" => finish_reason} when is_binary(finish_reason) ->
           %{
             "message" => %{"role" => "assistant", "content" => ""},
--- a/lib/req_llm/providers/google.ex
+++ b/lib/req_llm/providers/google.ex
@@ -1444,10 +1444,13 @@ defmodule ReqLLM.Providers.Google do
                   }
               end
 
+            response_with_raw_finish =
+              put_raw_finish_provider_meta(response_with_grounding, body)
+
             merged_response =
               ReqLLM.Context.merge_response(
                 req.options[:context] || %ReqLLM.Context{messages: []},
-                response_with_grounding
+                response_with_raw_finish
               )
 
             {req, %{resp | body: merged_response}}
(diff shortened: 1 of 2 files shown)


gi11es:
Heads-up on the two red checks — they are **pre-existing on `main`, not from this PR**:

- `CI / Quality / Docs` fails with `** (File.Error) could not read file "ROADMAP.md": no such file or directory`
- `CI / Quality / Hex Package Dry Run` fails for the same reason

`420fbbb9` ("docs: refresh guides and remove old roadmaps") deleted `ROADMAP.md`, `ROADMAP_V1.md` and `ROADMAP_V2.md`, but they are still referenced in `mix.exs` — in the `extras` list (lines 39-41 and 93-95) and in the hex `files` list (line 279). The CI run on `main` at 2026-08-01T14:13 is red for the same two jobs.

This branch is `main` plus one commit touching only `lib/req_llm/providers/google.ex` and `test/providers/google_test.exs`, and locally `mix test` (4051 passed), `mix quality` (dialyzer + credo --strict clean) and `mix mc "google:*"` are all green.

Worth fixing regardless of this PR, since the hex dry-run failure would block cutting a release. Happy to send a separate one-liner dropping the three stale references if that's useful.

PR #644, opened 2026-04-24
Feature/azure gpt image support

mrdotb:
## Description

Brief description of changes.

## Type of Change

- [ ] Bug fix (non-breaking change fixing an issue)
- [ ] New feature (non-breaking change adding functionality)
- [ ] Breaking change (fix or feature causing existing functionality to change)
- [ ] Documentation update

## Testing

- [ ] Tests pass (`mix test`)
- [ ] Quality checks pass (`mix quality`)

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have updated the documentation accordingly
- [ ] I have added tests that prove my fix/feature works
- [ ] All new and existing tests pass
- [ ] My commits follow conventional commit format
- [ ] I have **NOT** edited `CHANGELOG.md` (it is auto-generated by git_ops)

## Related Issues

Closes #

PR #956, opened 2026-08-10
feat: Azure image support

vasspilka:
## Description

Adding support for Azure OpenAI image models. 

Along the way we did some small refactoring to reuse some code and fixed a minor bug related to azure errors.

Note that Azure Foundry has it's own API that tries to unite different types of image model with one API. In this PR we focus solely on the OpenAI models but other providers should be supported (maybe with a tiny amount of additional work)

## Type of Change

- [ ] Documentation update

## Checklist

## Related Issues

Seems there was a related PR in the past https://github.com/agentjido/req_llm/pull/644 but was closed and never reopened


test/provider/azure/azure_test.exs
```diff
@@ -127,6 +127,26 @@ defmodule ReqLLM.Providers.AzureTest do
...
+                 :embedding,
+                 model,
+                 "Hello",
+                 deployment: "my-embedding-deployment",
+                 base_url: "https://my-resource.openai.azure.com/openai",
+                 provider_options: [dimensions: "not-an-integer"]
+               )
+
+      assert Exception.message(error) =~ "dimensions"
+    end
```
vasspilka, line 149: This test covers a fix we discovered while building the feature

vasspilka:
@mikehostetler Thanks for such a quick review :)

I discovered a few refactoring opportunities:

```
  1. The shared codec is in the wrong module. Azure fully-qualifies ReqLLM.Providers.OpenAI.ImagesAPI eight times. The reuse is right; the location makes Azure depend on OpenAI's namespace. The repo already has per-provider images_api.ex (xai, minimax), so a ReqLLM.Images.OpenAICompatible with OpenAI's ImagesAPI as a thin @behaviour adapter would be the truer shape. Not blocking — but it's what will look wrong the day a third provider reuses it.

  2. normalize_options/2 duplicates resolution that translate_options/2 already did. Both providers route through Options.process → translate_options, so by the time normalize_options/2 runs, :aspect_ratio is always either gone or malformed — the {:ok, resolved, _warnings} branches at images_api.ex:240-244 are dead in practice. The same nearest-size table now has two entry points and two sets of tests asserting it. normalize_options should only reject; resolution belongs in one place.

  3. The two-phase split is subtly coupled. translate_options deliberately leaves a malformed ratio untouched so normalize_options can hard-error it. That contract exists only in a comment. One function returning {opts, warnings} | {:error, _} would be simpler.

  4. @request_option_keys uses a denylist. Deriving from ReqLLM.Images.schema() is right; subtracting a hardcoded @non_request_schema_keys inverts the safety. A new plumbing key added to the schema silently becomes a registered request option. Blast radius is small (build_generation_body reads only known keys), but it fails silently. An allowlist — or a marker in the schema itself — would fail loudly.
 ```
 
 They are not critical. Will push the changes in a new branch and share a link to the commit. 
 We can proceed without them or include them to make the code slightly cleaner

mikehostetler:
@vasspilka My agents are fast :-)

I'll wait - I'm about to step away anyways and never merge without human review.

I'll take another pass this evening when you're ready and we can get this landed.

Thanks for the PR!

vasspilka:
Ok, pushed it. It basically extracts the OpenAI.ImagesAPI so it's more easily reusable with different providers, in case future providers use a similar API

PR #958, opened 2026-08-12
fix(streaming): fall back after oversized websocket messages

dannote:
## Description

OpenAI Responses WebSocket requests are sent as a single `response.create` text message. For large conversations, the upstream endpoint can reject that message with WebSocket close code `1009` (message too big). Retrying the same WebSocket payload cannot succeed and currently leaves callers with a transport error, even though the same request can be streamed through HTTP/SSE.

This PR adds a narrow, provider-declared fallback:

- recognize close code `1009` before any provider output has been emitted;
- replay the request once through the provider's existing HTTP/SSE streaming path;
- never replay after observable output, avoiding duplicate output or side effects;
- mark caller-owned reusable Responses WebSocket sessions as HTTP-only after fallback, so later turns do not repeat the known-failing transport;
- reset `StreamServer` protocol and provider state when switching from WebSocket events to SSE;
- reuse the existing Finch request construction, retry, telemetry, backpressure, and error handling paths.

Both OpenAI Responses and OpenAI Codex opt into the generic fallback through WebSocket request metadata. Other WebSocket providers retain their existing behavior.

This follows the transport strategy used by the official Codex client:

- [Responses WebSocket endpoint](https://github.com/openai/codex/blob/4ef836f883c38ba6d39e6920f335ce6452b7de33/codex-rs/codex-api/src/endpoint/responses_websocket.rs) sends `response.create` over WebSocket and enables `permessage-deflate`.
- [Responses retry orchestration](https://github.com/openai/codex/blob/4ef836f883c38ba6d39e6920f335ce6452b7de33/codex-rs/core/src/responses_retry.rs) switches transports after the WebSocket retry budget is exhausted.
- [Client transport state](https://github.com/openai/codex/blob/4ef836f883c38ba6d39e6920f335ce6452b7de33/codex-rs/core/src/client.rs) keeps fallback state at session scope.
- [Fallback integration tests](https://github.com/openai/codex/blob/4ef836f883c38ba6d39e6920f335ce6452b7de33/codex-rs/core/tests/suite/websocket_fallback.rs) verify HTTP fallback and that it remains sticky across turns.

ReqLLM intentionally implements the smallest corrective subset here: pre-output `1009` fallback and sticky routing. WebSocket compression and incremental `previous_response_id` reuse remain separate concerns.

## Type of Change

- [ ] Documentation update

## Breaking Changes

None.

## Testing

- [x] Tests pass (`mix test`)

Added integration coverage for:

- pre-output close `1009` replaying successfully over HTTP/SSE;
- close `1009` after provider output remaining an error without replay;
- reusable session fallback remaining sticky across subsequent turns;
- OpenAI Codex WebSocket requests declaring HTTP fallback support.

Validation performed:

- full suite: 4,136 passed, 11 skipped, 151 excluded;
- `mix quality`: formatting, warnings-as-errors compilation, Credo, and Dialyzer passed;
- real `openai_codex:gpt-5.6-sol` normal WebSocket stream succeeded;
- locally forced pre-output close `1009` then replayed successfully against the real Codex HTTP/SSE endpoint;
- the same smoke test verified sticky HTTP routing for the reusable session.

The real smoke test refreshed OAuth credentials in memory only and did not persist or log them.

## Checklist

- [x] Documentation is unchanged because this is automatic transport recovery with no new public option
- [x] I have added tests that prove my fix works

## Related Issues

No linked issue.

### DIFF
--- a/lib/req_llm/providers/openai.ex
+++ b/lib/req_llm/providers/openai.ex
@@ -805,7 +805,7 @@ defmodule ReqLLM.Providers.OpenAI do
   defp websocket_session_fell_back?(session) when is_pid(session) do
     ReqLLM.Streaming.WebSocketSession.http_fallback?(session)
   catch
-    :exit, _reason -> true
+    :exit, _reason -> false
   end
 
   defp websocket_session_fell_back?(_session), do: false
--- a/lib/req_llm/providers/openai.ex
+++ b/lib/req_llm/providers/openai.ex
@@ -791,13 +791,24 @@ defmodule ReqLLM.Providers.OpenAI do
 
   def stream_transport(_model, opts) do
     provider_opts = Keyword.get(opts, :provider_options, [])
+    session = Keyword.get(provider_opts, :openai_websocket_session)
 
     case Keyword.get(provider_opts, :openai_stream_transport, :sse) do
-      :websocket -> :websocket
-      "websocket" -> :websocket
-      _ -> :http
+      transport when transport in [:websocket, "websocket"] ->
+        if websocket_session_fell_back?(session), do: :http, else: :websocket
+
+      _ ->
+        :http
     end
   end
+
+  defp websocket_session_fell_back?(session) when is_pid(session) do
+    ReqLLM.Streaming.WebSocketSession.http_fallback?(session)
+  catch
+    :exit, _reason -> true
+  end
+
+  defp websocket_session_fell_back?(_session), do: false
 
   @doc false
   def resolve_request_credential!(%LLMDB.Model{} = model, opts) do
--- a/lib/req_llm/streaming/web_socket_client.ex
+++ b/lib/req_llm/streaming/web_socket_client.ex
@@ -33,7 +33,8 @@ defmodule ReqLLM.Streaming.WebSocketClient do
 
       :no_fixture ->
         with {:ok, config} <- build_stream_request(provider_mod, model, context, opts),
-             {:ok, task_pid} <- start_streaming_task(config, stream_server_pid, opts) do
+             {:ok, task_pid} <-
+               start_streaming_task(config, stream_server_pid, provider_mod, model, context, opts) do
           {:ok, task_pid, config.http_context, config.canonical_json}
         end
     end
@@ -53,7 +54,8 @@ defmodule ReqLLM.Streaming.WebSocketClient do
          headers: config.headers || [],
          initial_messages: config.initial_messages || [],
          http_context: http_context,
-         canonical_json: config.canonical_json || %{}
+         canonical_json: config.canonical_json || %{},
+         fallback_transport: Map.get(config, :fallback_transport)
        }}
     else
       false ->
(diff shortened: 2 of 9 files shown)


lib/req_llm/streaming/web_socket_client.ex
```diff
@@ -102,19 +104,33 @@ defmodule ReqLLM.Streaming.WebSocketClient do
...
-              initial_messages: config.initial_messages,
-              close_on_terminal?: false
-            )
-        end
+        session = reusable_session(opts)
+
+        result =
+          case session do
+            nil ->
+              start_owned_session(config, stream_server_pid, opts)
```
mikehostetler, line 115: [P1] Close an owned session on every terminal path. `start_owned_session/3` returns only the relay result, so the caller loses the session PID. After a code 1009 close, the task can replay over HTTP and exit while the owned `WebSocketSession` process remains alive. A deterministic process scan after the new test file found two such live sessions. Please retain the owned PID and stop that session after fallback or error. Do not stop a caller-supplied reusable session.

lib/req_llm/streaming/web_socket_client.ex
```diff
@@ -198,11 +218,101 @@ defmodule ReqLLM.Streaming.WebSocketClient do
         :ok
 
       {:error, reason} ->
-        safe_http_event(stream_server_pid, {:error, reason})
-        {:error, reason}
+        if message_too_big?(reason) do
+          {:websocket_error, reason, output_observed?}
```
mikehostetler, line 226: [P1] Preserve this error when HTTP fallback is not enabled. Every code 1009 close returns `{:websocket_error, reason, false}`, but the next function only handles that result when `fallback_transport` is `:http`. The generic clause then returns the tuple without sending an error event. The task exits normally, so `StreamServer` can finalize an empty successful stream. Please add an explicit no-fallback error path and a test for a provider or configuration with no `fallback_transport`.

lib/req_llm/providers/openai.ex
```diff
@@ -791,14 +791,25 @@ defmodule ReqLLM.Providers.OpenAI do
...
+
+      _ ->
+        :http
     end
   end
 
+  defp websocket_session_fell_back?(session) when is_pid(session) do
+    ReqLLM.Streaming.WebSocketSession.http_fallback?(session)
+  catch
+    :exit, _reason -> true
```
mikehostetler: [P2] Do not treat a dead or unresponsive reusable session as an established HTTP fallback. This catch changes the request to HTTP even though the session never set `http_fallback?`. That hides a caller-owned session lifecycle failure. Please select HTTP only when the live session reports `true`; otherwise keep the WebSocket path or return the session error. The same catch exists in `OpenAICodex`.

mikehostetler (changes requested):
Thanks for the focused fallback change and the useful transport tests. I found three blocking edge cases:

- An owned WebSocket session remains alive after an HTTP fallback or another error path.
- A code 1009 close becomes an empty successful stream when the provider does not enable HTTP fallback.
- A dead caller-owned session silently selects HTTP.

I opened #959 for the unrelated CI test-isolation failure. That race predates this PR and must stay separate from this change.

Local validation passed for the changed WebSocket and Codex tests, the streaming core tests, and the full suite with the CI seed. The current red CI result is the separate #959 race.

lib/req_llm/streaming/web_socket_client.ex
dannote: Fixed in 8bf3750a. Owned sessions are now closed after every relay terminal path, while caller-owned reusable sessions remain under caller ownership. Added a regression that verifies no owned WebSocketSession remains after HTTP fallback.

lib/req_llm/providers/openai.ex
dannote: Fixed in 8bf3750a for both OpenAI and OpenAICodex. A dead or unresponsive reusable session now reports false for sticky fallback selection, keeping the WebSocket path so the lifecycle failure remains observable. Added direct coverage for both providers.

lib/req_llm/streaming/web_socket_client.ex
dannote: Fixed in 8bf3750a. Any code 1009 result that does not match the provider-declared HTTP fallback clause now emits the original stream error instead of allowing a normal empty completion. Added a no-fallback provider regression.

mikehostetler (approved):
Thank you for the focused follow-up. Commit 8bf3750a resolves all three blocking findings and adds direct regression tests for each case. The focused tests, full local suite, merge check, and GitHub CI all pass. This is ready to merge.

PR #966, opened 2026-08-18
fix(openai): send null content for tool-only assistant messages

gilbertwong96:
## Description

DeepSeek, xAI, Groq and Mistral reject assistant messages that carry `tool_calls` but an empty-string `content` with `400: Messages with role 'tool' must be a response to a preceding message with 'tool_calls'`. The OpenAI spec says `content` is nullable when `tool_calls` is present, so the OpenAI-compatible encoder now sends `null` instead of `""` for messages that have `tool_calls` and no visible text (empty text parts or thinking-only content are treated as empty).

Adds `empty_content?/1` and updates the existing encoder tests that asserted the old `""` behavior.

## Type of Change

## Testing

- [x] Tests pass (`mix test`) — 4220 passed

## Checklist

### DIFF
--- a/lib/req_llm/provider/defaults.ex
+++ b/lib/req_llm/provider/defaults.ex
@@ -868,9 +868,19 @@ defmodule ReqLLM.Provider.Defaults do
        ) do
     {reasoning_content, content_without_thinking} = extract_reasoning_content(c)
 
+    # DeepSeek / xAI / Groq / Mistral reject tool-only assistant messages
+    # whose content is an empty string: the OpenAI spec says content is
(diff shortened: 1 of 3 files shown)


gilbertwong96:
Closed: verified against DeepSeek direct API and Bedrock OpenAI-compatible endpoint — both accept content:"" for tool-only assistant messages (200 OK). The premise (providers reject empty-string content) does not hold for these endpoints; the 400 the fix targeted was actually caused by tool_calls being dropped during loose-map conversion (fixed client-side).

PR #978, opened 2026-08-28
feat(bedrock): support application inference profiles

adriencohen:
Adds support for Amazon Bedrock [application inference profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/cost-mgmt-application-inference-profiles.html), which are addressed by an ARN rather than a model id. 

Two ways to use a profile:

- **ARN as the model id** (`amazon_bedrock:arn:aws:bedrock:…`): it names no family, so the request goes through the Converse API as one encoded path segment. No pricing is known.
- **`provider_options: [inference_profile_arn: arn]`** on a declared model the ARN replaces the id in the request path while the model spec keeps supplying family, capabilities, and pricing.

A native-format 400 sent through a profile now returns a hint naming the model the body was built for, since Bedrock only answers "schema violations" when the profile serves a different model. Any other model id is sent exactly as before.

## Type of Change

- [x] Documentation update

## Testing

- [x] Tests pass (`mix test`)

## Checklist

### DIFF
--- a/guides/amazon_bedrock.md
+++ b/guides/amazon_bedrock.md
@@ -23,6 +23,20 @@ For the full model-spec workflow, see [Model Specs](model-specs.md).
 
 Use exact Bedrock IDs from [LLM Catalog](https://llmcatalog.dev) when possible. The canonical ReqLLM provider prefix is `amazon_bedrock:`. For inference profiles, custom deployments, or new Bedrock model IDs, use a full explicit model spec when the registry has not caught up yet.
 
+An [application inference profile](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-create.html) is addressed by its ARN. Name the model it serves and pass the ARN as a provider option, so the request goes through the profile while capabilities and pricing come from the model:
+
+```elixir
+ReqLLM.generate_text(
+  "amazon_bedrock:anthropic.claude-sonnet-4-5-20250929-v1:0",
+  "Hello",
+  provider_options: [inference_profile_arn: "arn:aws:bedrock:us-east-1:123456789012:application-inference-profile/abcdef123456"]
+)
+```
+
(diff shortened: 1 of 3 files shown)


mikehostetler:
Independent review result: changes required.

- **P1 — Preserve ARN delimiters in the request path.** The [ARN path encoder](https://github.com/agentjido/req_llm/blob/8ecd434f4cbc68681932f237f53e09bf881132c9/lib/req_llm/providers/amazon_bedrock.ex#L909-L915) uses `URI.encode(arn, &URI.char_unreserved?/1)`. This changes `:` and `/` to `%3A` and `%2F`. Bedrock rejects this encoded ARN with HTTP 400. This blocks Converse, Invoke, streaming, and embedding requests that use an application inference profile. The same defect is documented in [Vercel issue #14117](https://github.com/vercel/ai/issues/14117), and its fix preserves these delimiters in [PR #17410](https://github.com/vercel/ai/pull/17410).
- The new tests assert the incorrect encoded path, so they do not detect this failure.
- Nonblocking: the comments at lines 911-912 are inside a function body, which conflicts with the repository instructions.

Validation: the PR merges cleanly with current `main`. The full merged test run completed with 4,255 passed, 11 skipped, and 152 excluded. CI is green, but the tests currently preserve the incorrect path behavior.

adriencohen:
Pushed a fix for 3.

Regarding 1 and 2, this is incorrect. The code was tested live, and the PR mentioned was reverted by https://github.com/vercel/ai/issues/14117  . The following is [still the code used in vercel/ai today](https://github.com/vercel/ai/blob/0a0f271cff061e1ca953ed9d14ed7d525613a4a9/packages/amazon-bedrock/src/amazon-bedrock-chat-language-model.ts#L1221):

```ts
return `${this.config.baseUrl()}/model/${encodeURIComponent(modelId)}`;
```

mikehostetler:
@adriencohen thanks - AWS Bedrock is one of the few providers that I'm not set up to test with a live request - will absorb and land this

PR #998, opened 2026-09-10
Fix max tokens usage for bedrock mantle

michaeljguarino:
## Description

Openai models force block the max_tokens field that mantle requests were defaulting for openai.* models.  This should convert it to the expected format, alongside some tests to confirm functionality.

The apparent root cause here is we need to detect more precisely what the friendliest openai endpoint to use for each model is (between chat completions and responses), and a lot of these fields naturally translate once that's settled.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Tests pass (`mix test`)

## Checklist

- [ ] My commits follow conventional commit format

## Related Issues

Closes #

---
<a href="https://app.blacksmith.sh/agentjido/codesmith/req_llm/pr/998"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791605606&installation_model_id=434874&pr_number=998&repository=agentjido%2Freq_llm&return_to=https%3A%2F%2Fgithub.com%2Fagentjido%2Freq_llm%2Fpull%2F998&signature=254614801f1b06925ecc55bb571e33ca6e2e3b392fae80a3163d0646c0beff89"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/lib/req_llm/providers/amazon_bedrock.ex
+++ b/lib/req_llm/providers/amazon_bedrock.ex
@@ -398,7 +398,7 @@ defmodule ReqLLM.Providers.AmazonBedrock do
     use_converse = determine_use_converse(model_id, opts)
 
     {endpoint_base, formatter, model_family} =
-      route(endpoint, model_id, use_converse, opts[:stream] == true, opts)
+      route(endpoint, model, use_converse, opts[:stream] == true, opts)
 
     operation = opts[:operation] || :chat
     compat_opts = Keyword.put(opts, :use_converse, use_converse)
@@ -425,7 +425,8 @@ defmodule ReqLLM.Providers.AmazonBedrock do
         :operation,
         :tools,
         :inference_profile_arn,
-        :endpoint
+        :endpoint,
+        :response_formatter
       ])
       |> Req.Request.merge_options(
         ReqLLM.Provider.Defaults.finch_option(request) ++
@@ -438,7 +439,8 @@ defmodule ReqLLM.Providers.AmazonBedrock do
             context: opts[:context],
             use_converse: use_converse,
             operation: opts[:operation],
-            tools: opts[:tools]
+            tools: opts[:tools],
+            response_formatter: formatter
           ]
       )
 
@@ -563,7 +565,7 @@ defmodule ReqLLM.Providers.AmazonBedrock do
     use_converse = determine_use_converse(model_id, translated_opts)
 
     {path, formatter, model_family} =
-      route(endpoint, model_id, use_converse, true, translated_opts)
+      route(endpoint, model, use_converse, true, translated_opts)
 
     translated_opts = Keyword.put(translated_opts, :use_converse, use_converse)
 
@@ -690,9 +692,12 @@ defmodule ReqLLM.Providers.AmazonBedrock do
   def decode_stream_event(%{data: _} = event, model, state) do
     model_id = model.provider_model_id || model.id
 
-    case mantle_wire(model_id) do
+    case mantle_wire(model) do
       :messages ->
         ReqLLM.Providers.Anthropic.Response.decode_stream_event(event, model, state)
+
+      :responses ->
+        ReqLLM.Providers.AmazonBedrock.Responses.decode_stream_event(event, model, state)
 
       :chat_completions ->
         openai = %{model | id: model_id, provider: :openai}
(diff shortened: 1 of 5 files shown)


michaeljguarino:
(happy to reword this as an issue if you'd prefer to implement yourself, but seemed trivial so thought I'd just do it)

lib/req_llm/providers/amazon_bedrock/responses.ex
```diff
@@ -0,0 +1,101 @@
...
+    opts
+    |> Map.new()
+    |> Map.merge(%{
+      model: openai_model.id,
+      id: openai_model.id,
+      context: context,
+      req_llm_model: openai_model
+    })
+    |> then(&%{options: &1})
+    |> ResponsesAPI.build_body()
```
mikehostetler, line 24: [P1] Encode the compiled object schema before delegation

This adapter calls the OpenAI Responses body builder directly, so it skips the native OpenAI object preparation step. For an object request, compiled_schema is not converted to text.format or to a strict tool. I reproduced a generated /responses body with neither field. This makes generate_object unconstrained for every Bedrock model that uses this route. Please apply the Responses object preparation logic and add a request-level object test.

lib/req_llm/providers/amazon_bedrock/responses.ex
```diff
@@ -0,0 +1,101 @@
...
+    |> Map.drop([:model, "model"])
+    |> Map.put(:model, model_id)
```
mikehostetler, line 26: [P1] Provide a way to disable response storage

The new Responses request omits store, and Amazon Bedrock Mantle stores Responses data by default. The Bedrock provider schema also rejects provider_options: [store: false], so ReqLLM users cannot opt out. This changes the data-retention behavior of these calls. Please default to store: false, or expose and validate the Responses storage option. AWS documents the default here: https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html

lib/req_llm/providers/amazon_bedrock.ex
```diff
@@ -690,10 +692,13 @@ defmodule ReqLLM.Providers.AmazonBedrock do
   def decode_stream_event(%{data: _} = event, model, state) do
     model_id = model.provider_model_id || model.id
 
-    case mantle_wire(model_id) do
+    case mantle_wire(model) do
       :messages ->
         ReqLLM.Providers.Anthropic.Response.decode_stream_event(event, model, state)
 
+      :responses ->
+        ReqLLM.Providers.AmazonBedrock.Responses.decode_stream_event(event, model, state)
```
mikehostetler, line 700: [P2] Use the Responses response builder for streams

This decoder emits response.completed with finish_reason :stop. The native OpenAI Responses ResponseBuilder changes that value to :tool_calls when the stream contains actionable tool calls. Bedrock models use the default builder, so I reproduced a materialized response with one tool call and finish_reason :stop. This can make an agent treat a tool turn as complete. Please select the Responses builder for this route, or normalize the finish reason in this adapter, and test the materialized stream result.

mikehostetler (changes requested):
I found three blocking regressions in the new Responses route. The added unit tests pass, but they do not cover object generation, storage controls, or final stream materialization.

lib/req_llm/providers/amazon_bedrock.ex
```diff
@@ -690,10 +692,13 @@ defmodule ReqLLM.Providers.AmazonBedrock do
   def decode_stream_event(%{data: _} = event, model, state) do
     model_id = model.provider_model_id || model.id
 
-    case mantle_wire(model_id) do
+    case mantle_wire(model) do
```
mikehostetler, line 695: [P1] Do not use the Mantle wire selector for runtime streams

This callback also handles `bedrock-runtime` streams, but it now chooses `:responses` only from the model. For an `openai.gpt-5.*` request without `endpoint: :mantle`, `route(:runtime, ...)` still formats `/invoke-with-response-stream` with `AmazonBedrock.OpenAI` (Chat Completions), while this branch sends the returned `choices[].delta` events to the Responses decoder. Those events produce no content chunks. Keep the runtime decoder on its selected formatter, or carry the selected endpoint and wire format into stream decoding. Add a runtime streaming test for a GPT-5 model.

PR #1004, opened 2026-09-11
feat(bedrock): send documents and video on the Converse API

adriencohen:
## Description

The Converse formatter dropped `file`, `image_url` and `video_url` parts, so a PDF or a video never reached the model. They now become the [document, image and video blocks](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html#converse-messages) of a Converse message, chosen by media type or, for the default `application/octet-stream`, by file extension. Sources are inline bytes or an `s3://` URL with optional `bucket_owner` metadata. A document is named after its `title` metadata or its filename, and `context` metadata passes through:

```elixir
ReqLLM.Context.user([
  ReqLLM.Message.ContentPart.text("What's in this document?"),
  ReqLLM.Message.ContentPart.file(pdf, "MyDocument.pdf", "application/pdf")
])
```

Formats are sent as given, so Amazon refuses what it does not accept, and unknown image media types no longer fall back to `png`. Sources Converse cannot fetch (http URLs, provider file ids) raise `ReqLLM.Error.Invalid.Parameter` instead of disappearing. 

## Type of Change

- [x] Documentation update

## Testing

- [x] Tests pass (`mix test`)
- [x] Fixture recorded live

## Checklist

---
<a href="https://app.blacksmith.sh/agentjido/codesmith/req_llm/pr/1004"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791731050&installation_model_id=434874&pr_number=1004&repository=agentjido%2Freq_llm&return_to=https%3A%2F%2Fgithub.com%2Fagentjido%2Freq_llm%2Fpull%2F1004&signature=829ad01508d79c09100a1d83cc6aeaed6cd272fb5913be18dac12e56853ed0be"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/lib/req_llm/providers/amazon_bedrock/converse.ex
+++ b/lib/req_llm/providers/amazon_bedrock/converse.ex
@@ -86,6 +86,26 @@ defmodule ReqLLM.Providers.AmazonBedrock.Converse do
   @reasoning_format "bedrock-converse-v1"
   @replayable_reasoning_providers [:amazon_bedrock, :anthropic]
 
+  @image_formats ~w(png jpeg gif webp)
+  @video_formats ~w(mkv mov mp4 webm flv mpeg mpg wmv three_gp)
+  @format_aliases %{
+    "jpg" => "jpeg",
+    "quicktime" => "mov",
+    "x-matroska" => "mkv",
+    "matroska" => "mkv",
+    "x-flv" => "flv",
+    "x-ms-wmv" => "wmv",
+    "3gpp" => "three_gp",
+    "3gp" => "three_gp",
+    "plain" => "txt",
+    "markdown" => "md",
+    "htm" => "html",
+    "msword" => "doc",
+    "vnd.openxmlformats-officedocument.wordprocessingml.document" => "docx",
+    "vnd.ms-excel" => "xls",
+    "vnd.openxmlformats-officedocument.spreadsheetml.sheet" => "xlsx"
+  }
+
   @doc """
   Format a ReqLLM context into Bedrock Converse API format.
 
@@ -487,11 +507,11 @@ defmodule ReqLLM.Providers.AmazonBedrock.Converse do
   end
 
   defp encode_system_message(%Message{content: content}) when is_binary(content) do
-    encode_content_for_system(content)
+    encode_content(content)
   end
 
   defp encode_system_message(%Message{content: content}) when is_list(content) do
-    encode_content_for_system(content)
+    encode_content(content)
   end
 
   defp encode_system_message(_message), do: []
@@ -728,14 +748,6 @@ defmodule ReqLLM.Providers.AmazonBedrock.Converse do
 
   defp encode_reasoning_detail(_detail), do: []
 
-  defp encode_content_for_system(content) when is_binary(content) do
-    [%{"text" => content}]
-  end
-
-  defp encode_content_for_system(content) when is_list(content) do
-    Enum.map(content, &encode_content_part/1)
-  end
-
   defp encode_content(content) when is_binary(content) do
     [%{"text" => content}]
   end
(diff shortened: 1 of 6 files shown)


lib/req_llm/providers/amazon_bedrock/converse.ex
```diff
@@ -752,18 +764,83 @@ defmodule ReqLLM.Providers.AmazonBedrock.Converse do
...
+  end
+
+  defp block_type(format) when format in @image_formats, do: "image"
+  defp block_type(format) when format in @video_formats, do: "video"
+  defp block_type(_format), do: "document"
+
+  defp media_block("document", part, format) do
+    source = encode_source(part)
+
+    %{"name" => document_name(part), "format" => format, "source" => source}
```
mikehostetler, line 886: [P2] Reject a document-only message

Amazon Bedrock requires a Message that contains a DocumentBlock to also contain a text block with a related prompt. This formatter accepts Context.user([ContentPart.file(...)]) and emits a document-only message, so AWS rejects the request. Please validate this at the message level and return ReqLLM.Error.Invalid.Parameter, or create another supported behavior. Add a request-level test for this case. AWS documents the rule here: https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html#converse-messages

lib/req_llm/providers/amazon_bedrock/converse.ex
```diff
@@ -752,18 +764,83 @@ defmodule ReqLLM.Providers.AmazonBedrock.Converse do
...
+
+  defp encode_source(%ContentPart{file_id: file_id}) when is_binary(file_id),
+    do: invalid_part("Converse cannot read provider file ids")
+
+  defp document_name(%ContentPart{filename: filename, metadata: metadata}) do
+    (metadata_value(metadata, :title) || filename_stem(filename))
+    |> String.replace(~r/[^A-Za-z0-9 \-()\[\]]+/, " ")
+    |> String.replace(~r/ +/, " ")
+    |> String.slice(0, 200)
+    |> String.trim()
```
mikehostetler, line 931: [P2] Keep the required document name nonempty

The cleanup can remove every character. For example, ContentPart.file("x", "!!!.pdf", "application/pdf") produces name: "". Amazon Bedrock requires a document name with 1 to 200 characters, so this accepted input always fails at the API. An S3 document without a title or filename has the same result. Please use a neutral nonempty fallback or raise ReqLLM.Error.Invalid.Parameter, and add tests. The constraint is here: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentBlock.html

mikehostetler (changes requested):
I found two request-shape defects in the new document path. The focused tests pass, but they do not cover these AWS validation rules.

mikehostetler (changes requested):
Re-check after conflict resolution at `39cb36eb`:

- The branch now merges cleanly with `main`. The combined guard-content and attachment paths pass 62 focused unit tests, 2 recorded AWS coverage tests, and the full local suite (4,352 passed; 11 skipped).
- Two request-shape blockers remain reproducible on the new head. A document-only message emits a `document` block without a related `text` block, which the Converse API does not accept. Also, the document-name sanitizer can emit `name: ""` (for example, `!!!.pdf`), but AWS requires a name length of 1 to 200 characters.

Please add validation or a safe fallback for both cases, with regression tests. The PR stays `needs_work`.

mikehostetler (approved):
Re-check after `3ccfb139`: the two AWS request-shape blockers are fixed. Document messages now require related text, empty sanitized names use the neutral `Document` fallback, and direct system attachments are rejected. The new regression tests, full local suite, AWS fixture tests, and all GitHub checks pass.

Issue #1015, opened 2026-09-16
[Feature]: A2Agent model specs and live compatibility fixture using OpenAI defaults


A2agent-ai:
### Feature Type

New provider

### Description

Hi ReqLLM maintainers — I represent [A2Agent](https://a2agent.me), an OpenAI- and Anthropic-compatible API gateway.

ReqLLM's provider guide says the default implementation is OpenAI-compatible and recommends adding a provider module only for actual deviations. We would therefore like to start with the smallest provider-neutral contribution:

- explicit A2Agent model specs or a configuration example, if that fits the registry policy;
- recorded fixtures for non-streaming and Finch-backed SSE streaming;
- compatibility checks for tools, usage normalization and structured provider errors; and
- an optional live test using a dedicated low-balance key, without exposing prompts or response bodies in logs.

Would you prefer explicit model specs, documentation only, or a generic external-gateway fixture? We would not create a provider module unless testing shows that the OpenAI-compatible defaults are insufficient.

Separately, if the technical evidence is accepted, would you consider a factual A2Agent link under **Tested Providers**, **Integrations**, or another location you choose? Please confirm the wording/location and whether a logo is acceptable. This request is optional, separate from technical review, and does not imply endorsement.

A2Agent can provide a private test key, technical/documentation support, and **$20–$100 in non-cash API testing credits**, with possible continuing credits for maintained compatibility tests. There is no referral or revenue-share condition, and the support is not consideration for a merge, endorsement, or displaying a link.

### Can you contribute this?

Yes, I can submit a PR

mikehostetler:
Thanks for the proposal. I opened #1017 for one shared ReqLLM path for gateways that use OpenAI Chat Completions. Your suggestion to start with model data fits that work.

Please open a PR to [LLMDB](https://github.com/agentjido/llmdb) with A2Agent's provider and model metadata, if it is not there already. Include the exact model IDs, API base URL, bearer API key environment variable, declared OpenAI-compatible operations, and accurate capability and pricing data with source links. The [LLMDB contribution guide](https://github.com/agentjido/llmdb/blob/main/CONTRIBUTING.md#provider-data-contributions) gives the data and build steps. Once the shared path is released and the LLMDB data is available, `a2agent:<model-id>` can use it. Anthropic-compatible behavior is outside the initial scope of #1017.

[...]

Issue #999, opened 2026-09-10
[Feature]: OrcaRouter provider support for ReqLLM


clementguarino06510-glitch:
**Feature Type:** New provider

**Provider:** orcarouter

**Description:**

ReqLLM tackles a hard problem — "LLM APIs are inconsistent" — with one idiomatic Elixir surface (`generate_text/3`, `stream_text/3`, `generate_object/4`) spanning 1,205 models and 21 provider integrations. The design choice behind that breadth stands out: OpenAI Chat Completions is treated as the baseline standard, so a provider that speaks it needs only minimal overrides via `ReqLLM.Provider.DSL`, and the `Default` implementation is already OpenAI-compatible. For Elixir teams, that keeps each new backend small and additive rather than a rewrite of their call sites.

That extensibility is why another provider is worth raising. Elixir teams often want one key that reaches many models and keeps working when a single upstream degrades — the need the existing `openrouter` provider already serves. OrcaRouter fits that same shape: an additional OpenAI-compatible option alongside it, not a replacement.

I'm an engineer on the OrcaRouter team. I'd like to propose **OrcaRouter as an optional provider**. It would be purely additive: it does not replace, wrap, or alter any existing provider, and current configurations would be untouched.

OrcaRouter exposes an OpenAI-compatible API and uses standard API-key authentication, so it maps onto ReqLLM's baseline path rather than needing a custom transport. The expected integration point looks like a module at `lib/req_llm/providers/orcarouter.ex` declaring `id`, `default_base_url`, and `default_env_key`, then reusing the OpenAI-style encode/decode defaults with light DSL overrides — the same shape as the existing `openrouter` provider. To be clear, I have not written or tested any code; this is a proposal for your consideration first.

Capabilities that seem most relevant here:

- Multiple chat, reasoning, image, and video models behind one endpoint and one key.
- Automatic model routing and provider failover, so a request can survive one upstream degrading — analogous to the routing options ReqLLM already exposes for OpenRouter.
- Usage tracking and budgets, which pair with ReqLLM's normalized `response.usage` and cost estimation.

OrcaRouter already appears across open-source projects including RAGFlow, Dify, goose, and promptfoo.

One transparent note: we run an optional open-source partner program in which approved OSS projects can receive a 5% revenue share from OrcaRouter usage attributed to their integration. Participating is entirely optional and is not a precondition for the integration — I'm raising this on technical merit, and I'm glad to follow ReqLLM's disclosure and governance expectations.

[...]

PR #1019, opened 2026-09-16
Add shared OpenAI-compatible adapter for community providers

### LINKED ISSUE #1017: Add generic OpenAI-compatible dispatch for LLMDB catalog providers
## Problem

LLMDB can catalog models under a gateway's own provider ID and record that its API uses OpenAI Chat Completions. ReqLLM still requires a registered provider module with that exact ID before it can call a model. This leads to a small provider module for each gateway even when the shared OpenAI-compatible request and response code is sufficient.

This affects the requests in #999 (OrcaRouter) and #1015 (A2Agent).

## Proposed behavior

Add a model-aware generic adapter for OpenAI Chat Completions-compatible catalog entries. For example, if LLMDB contains `orcarouter:<model-id>` with the required runtime metadata, ReqLLM should be able to call it while preserving `:orcarouter` as the model's provider ID.

- Use an existing registered provider module first. Use the generic adapter only when no module is registered for the model's provider ID.
- Require explicit LLMDB execution metadata for the requested operation, plus a usable base URL and supported authentication metadata. Do not infer compatibility from the provider name or from catalog membership alone.
- Use the provider's model ID, endpoint, and credentials. Keep usage, errors, and telemetry attributed to the catalog provider ID.
- Reuse `ReqLLM.Provider.Defaults` for the OpenAI Chat Completions request, response, and SSE formats. Do not route a gateway model through OpenAI's Responses API based on its model name.
- Return a clear error for a missing runtime contract or an operation that the catalog does not declare as compatible.
- Keep provider-specific modules available for APIs that need different request or response behavior.

A module with only the fixed ID `:openai_compatible` is not enough: a catalog model still has its own provider ID, and current provider lookup and the default request code require a matching module ID. The dispatch path must account for that identity.

## Initial scope and checks

Start with text generation, structured output, and streaming through the OpenAI Chat Completions surface. Add deterministic tests for generic dispatch, existing-module precedence, provider-specific credentials and URL selection, and unsupported-operation errors. Named gateways that use this path have the **community supported** label. Their contributors maintain LLMDB model metadata and report compatibility problems. ReqLLM will use deterministic tests for the shared adapter. Maintainers will not run live fixture tests for each community gateway. A catalog entry identifies the declared API format; it does not verify every model feature.

## Related issues

- #999 — OrcaRouter provider request
- #1015 — A2Agent model specs and compatibility request

mikehostetler:
## Summary

- Route LLMDB models with an executable OpenAI Chat Completions contract through a shared adapter when they have no dedicated ReqLLM provider module.
- Use the provider URL and API key environment variable from LLMDB. Keep dedicated provider modules in control when they exist.
- Document the Community supported policy. These providers need correct LLMDB metadata and do not receive live ReqLLM fixtures.

## Tests

- `mix test test/req_llm test/providers` — 3660 passed, 11 skipped, 25 excluded
- `mix format --check-formatted`
- `mix compile --warnings-as-errors`

Tests use mock HTTP responses. No live provider calls were made.

Closes #1017. Related requests: #999 and #1015.

---
<a href="https://app.blacksmith.sh/agentjido/codesmith/req_llm/pr/1019"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792173744&installation_model_id=434874&pr_number=1019&repository=agentjido%2Freq_llm&return_to=https%3A%2F%2Fgithub.com%2Fagentjido%2Freq_llm%2Fpull%2F1019&signature=04ca369a09f64062fd7b873048e9f171631386a0440e46d23a073e10affffbf5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


test/providers/openai_compatible_test.exs
```diff
@@ -0,0 +1,147 @@
+defmodule ReqLLM.Providers.OpenAICompatibleTest do
+  use ExUnit.Case, async: true
+
+  alias ReqLLM.Providers.OpenAICompatible
+
+  defp model(attrs \\ %{}) do
```
ccarvalho-eng: we could move this to the bottom of the module or to a fixtures/* module

mikehostetler:
Closing in favor of #1028, which implements the same catalog-gateway goal with the current execution-metadata design and passing CI.

PR #1024, opened 2026-09-17
fix(anthropic): replay tool search blocks in place in the assistant turn

vildeserina:
Follow-up to #1013, which added Anthropic tool search but left the search-result blocks dropped on decode. The PR describes the cost as "one in-request re-search per turn rather than an error". With extended thinking on it is an error.

## What happens today

The decoder maps `server_tool_use` and `tool_search_tool_result` to `decode_content_block(_) -> nil`, so the replayed assistant turn loses them. The API compares that turn against what it produced and rejects it:

```
400 invalid_request_error
messages.1.content.1: `thinking` or `redacted_thinking` blocks in the latest
assistant message cannot be modified. These blocks must remain as they were
in the original response.
```

Any tool loop running a deferred catalog on a thinking model fails on its second round. `ReqLLM.Error.API.Stream` is raised rather than returned, so the caller's task crashes.

## Measured against the live API

Round 1 on `claude-opus-5` with a deferred tool catalog returns

```
thinking · text · server_tool_use ×2 · tool_search_tool_result ×2 · thinking · tool_use
```

Replaying that turn three ways:

| replayed turn | result |
|---|---|
| unchanged | 200 |
| server blocks dropped (ReqLLM today, no thinking) | 200, at the cost of a re-search |
| thinking hoisted to the front (ReqLLM today, thinking on) | **400** |

So it is not enough to keep the blocks: the trailing thinking block has to keep its place, and `encode_message/1` hoists every thinking block to the front of the turn.

## The change

* New `ContentPart` type `:provider_block`, holding the raw block plus the owning provider, built with `ContentPart.provider_block/3`.
* `Anthropic.Response` decodes `server_tool_use` and `tool_search_tool_result` into it, buffered and streaming. In the stateful stream decoder a `server_tool_use` block is held until `content_block_stop` so its `input_json_delta` fragments assemble into the block instead of leaking as orphan tool-call argument fragments; `tool_search_tool_result` arrives whole.
* A thinking block that **follows** a server block decodes the same way, because only the ordered content keeps its position. Thinking before the first server block keeps the reasoning-detail path, so a turn without server tools decodes and encodes byte-for-byte as before.
* `Anthropic.Context` replays a block it owns verbatim at its position and drops blocks another provider owns.
* `usage.server_tool_use.tool_search_requests` is surfaced next to the web search/fetch counts.

## Verification

* `mix test` — 4496 tests, 0 failures, 11 skipped. `mix quality` passes (format, `--warnings-as-errors`, credo --strict, dialyzer).
* Six new tests in `test/providers/anthropic_test.exs` under `describe "tool search round-trip"`, including a replay assertion on the encoded turn and a check that a turn without server tools is unaffected.
* End to end through `AshAi.ToolLoop` on `claude-opus-5` with 263 deferred tools: search → catalog tool → write tool, three iterations, no 400. The same loop fails on round 2 with v1.24.0.

---
<a href="https://app.blacksmith.sh/agentjido/codesmith/req_llm/pr/1024"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792261239&installation_model_id=434874&pr_number=1024&repository=agentjido%2Freq_llm&return_to=https%3A%2F%2Fgithub.com%2Fagentjido%2Freq_llm%2Fpull%2F1024&signature=0ccb9aacb36e92e89e7c17d5fd38bcd2ae2c71305f25bf3e2e82fb91a0f7f832"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


mikehostetler:
@vildeserina Could a `kult-byra` admin grant `@mikehostetler` write access to `kult-byra/req_llm`? GitHub rejected my push to the `tool-search-roundtrip` branch because I do not have write access to the organization-owned fork.

The patch for the three review findings is in [commit 7c5aeba1](https://github.com/agentjido/req_llm/commit/7c5aeba12841763dd245088536c9e1922f4e5ec4). The full local test suite and `mix quality` pass. With write access, I can push it to this PR branch.

mikehostetler:
Direct push access to `kult-byra/req_llm:tool-search-roundtrip` is still denied. I prepared a mergeable helper PR with the validated follow-up commits: https://github.com/kult-byra/req_llm/pull/1

It includes commit `7c5aeba1` plus `1d2423ab`, which preserves `redacted_thinking` blocks and makes the new provider-block content type safe to inspect.

Local validation passed:

- `mix test test/providers/anthropic_test.exs test/req_llm/messaging_test.exs` — 121 passed
- `mix quality` — passed

Exact next action: merge kult-byra/req_llm#1 into `tool-search-roundtrip`. That will update this PR and start fresh CI. This PR remains labeled `needs_work` until that happens and the new checks are green.

mikehostetler:
Superseded by #1034, which contains the original change and all validated follow-up fixes on an upstream branch.

PR #1025, opened 2026-09-18
fix(deepseek): map reasoning effort levels to supported values

mattneel:
## Summary

DeepSeek's reasoning-effort translation now preserves the requested level instead of collapsing multiple levels to the wrong API value.

This PR maps the requested effort levels as follows:

| ReqLLM effort | DeepSeek API value |
| --- | --- |
| `:minimal` | `"low"` |
| `:low` | `"low"` |
| `:medium` | `"high"` |
| `:high` | `"high"` |
| `:xhigh` | `"high"` |
| `:max` | `"max"` |
| `:ultra` | `"max"` |

It also preserves the existing `:none` and `:default` handling, keeps string values unchanged, and verifies the encoded request body rather than only the intermediate options.

DeepSeek documents `none`, `low`, `high`, and `max` as supported reasoning-effort values:
https://api-docs.deepseek.com/api/create-chat-completion/

## Testing

- `mix test test/providers/deepseek_test.exs` — 29 tests, 0 failures
- `mix format --check-formatted` — passed
- `mix compile --warnings-as-errors` — passed
- `mix credo --strict` — passed (`9144 mods/funs, found no issues`)
- `mix test` — 4492 tests, 0 failures, 11 skipped (193 excluded)
- `mix quality` — format, compile, and Credo passed; the command exits at Dialyzer on the existing unrelated warning at `lib/req_llm/streaming/http2_duplex_session.ex:83`


mikehostetler:
I found one validation gap: `translate_options/3` mapped `:ultra` to `"max"`, but the shared option schema rejected `:ultra` before translation.

I pushed commit `0c7731be` to map `:ultra` to `:max` in DeepSeek's pre-validation step. The new test checks that `prepare_request/4` accepts `:ultra` and encodes `"max"`. The shared schema stays unchanged for other providers.

Verification: 30 DeepSeek tests passed, formatting and compile checks passed, and a live `deepseek-flash` request using `:ultra` returned HTTP 200 with `"max"` in the request body. All required GitHub checks passed on the new commit, and the PR is mergeable.

